PR #120, opened 2026-06-27
feat(backend): add one-time HMAC locker provisioning

### LINKED ISSUE #110: New Feature: reset provisioning token from filament admin panel

Aibolit66:
Closes #110.
Implements #159.

## Goal

Provide an administrative locker-provisioning recovery flow without storing
reusable plaintext provisioning credentials in the database or event store.

This remains an administrative recovery workflow, not complete secure
revocation of an already established device identity.

## Provisioning behavior

- Delegate issue/restart orchestration to `LockerProvisioningService` and enforce
  `lockerbank.configure` inside the service.
- Generate a one-time token, store only its HMAC-SHA-256 digest, and return the
  plaintext token only to the initiating Filament request.
- Guard queued/replayed provisioning work with a per-attempt generation UUID so
  stale work cannot consume or complete a newer attempt.
- Record locker-bank UUID, initiating actor, generation, and issue timestamp in
  the audit event without token material.
- Delete the existing MQTT user, mark the bank offline, clear heartbeat and
  configuration sent/ack state, and preserve compartments and domain history.
- Redact token-bearing registration topics in backend and locker-client logs.

## Key derivation and migration

The provisioning HMAC key is not configured separately and `APP_KEY` is not used
directly. The backend validates Laravel's `APP_KEY`, decodes its raw or `base64:`
form, and derives a 32-byte subkey with HKDF-SHA-256 using the versioned domain
`open-locker/provisioning-token-hmac/v1`.

The migration removes the legacy plaintext token column and invalidates all open
legacy tokens. Operators must issue a new one-time token for unprovisioned
lockers after deployment. Rotating `APP_KEY` likewise invalidates outstanding,
unconsumed tokens, while already provisioned clients retain their MQTT
credentials.

## Admin one-time flow

- Issue or restart provisioning from the locker-bank list or edit page.
- Automatically open the copy modal after confirmation; there is no separate
  visible show action.
- Keep the plaintext token out of Livewire snapshots and session state, clear it
  when the modal closes, and prevent reopening it.
- Use consistent English and German labels and the `Issued` / `Erstellt` status.

## Explicit scope boundary

This PR deletes the legacy UUID MQTT user during reset and recreates that
username only after successful provisioning. It does not guarantee immediate
termination of an established MQTT TCP session or introduce unique
per-provisioning MQTT identities, broker eviction, retained-message migration,
or durable denial after username recreation. Those lifecycle guarantees remain
tracked in #161.

## Architecture decision

See [`docs/adr/0034-one-time-hmac-locker-provisioning.md`](https://github.com/Open-Locker/Open-Locker/blob/feat/110-reset-provisioning-token/docs/adr/0034-one-time-hmac-locker-provisioning.md).
The decision is **Accepted**.

## Verification

- Backend: focused provisioning and registration suites — 26 tests, 144
  assertions.
- Locker client: TypeScript build and 4 focused MQTT transport tests.
- Merge hooks: Laravel Pint; mobile typecheck, lint, and format check.
- IDE diagnostics: no errors in the conflict-resolved files.


merona-merlin:
## Review findings: agreed scope for provisioning reset

The aggregate → event → projector flow is directionally consistent, but the reset needs the following security and lifecycle changes before merge. This PR remains an administrative recovery workflow; durable MQTT credential-generation management is deferred to #161.

### 1. Keep the current token workflow, but do not persist the token in the event store

For this PR, keep the existing `locker_banks.provisioning_token` behavior and the current admin workflow for copying the token. Do **not** introduce HMAC storage or encryption as part of #120.

`LockerProvisioningReset` must not contain `newProvisioningToken`. The dedicated reset service should generate and write the new token to the locker-bank row, while the event records audit metadata only: locker-bank UUID, initiating actor, and reset timestamp.

This intentionally means provisioning tokens remain outside event replay. That is consistent with current behavior: locker banks are created directly through Eloquent and `LockerBank::booted()` generates the initial token in a `creating` hook, so a clean replay cannot reconstruct initial tokens today.

The broader HMAC question is deferred to #159.

### 2. Record the actor and timestamp

The PR description promises an audit trail for “who reset which device, when”, but the event contains no actor. Add at least `actorUserId` and `resetAt`, and register the event in the audit-log presenter without rendering token material.

### 3. Move orchestration and authorization into a service

The Filament action calls `LockerBankAggregate` directly and authorization is represented by action visibility. Introduce a focused provisioning-reset service that:

- resolves/receives the authenticated actor;
- enforces `lockerbank.configure` server-side;
- performs token rotation, read-model cleanup, and MQTT-user deletion;
- records the actor-aware event without the token.

The Filament resource should only call that service. Add an unauthorized direct-service test, not only an action-visibility assertion.

### 4. Delete the existing MQTT user, with an explicit scope boundary

Reset must delete the existing MQTT user identified by the locker-bank UUID. `MqttUserService::deleteUser()` already supports this. With `auth_opt_cache false`, subsequent authentication and ACL checks for the missing user are denied by Laravel.

This PR does **not** guarantee that Mosquitto immediately closes an already-established TCP session. It also does not solve the possibility that an old session regains ACL rights when the same UUID username is recreated during re-provisioning.

Do not add an application-level “reject all messages while unprovisioned” guard in this PR. Such a guard protects only the temporary null-`provisioned_at` window and does not solve username reuse after re-provisioning.

Unique per-provisioning MQTT identities, migration of existing credentials, durable denial of old sessions, retained-message behavior, and optional broker eviction are tracked in #161.

Tests in this PR should prove that:

- the old MQTT user row is removed during reset;
- the auth and ACL endpoints deny a deleted user;
- successful re-provisioning recreates the same legacy username format (locker-bank UUID) with a newly generated password.

A live-broker test for old-session behavior belongs to #161 because that issue changes the identity lifecycle that makes the guarantee durable.

### 5. Mark the locker bank offline

A reset invalidates the current device state. Reset at least:

- `connection_status` to `offline`;
- `connection_status_changed_at` to the reset timestamp;
- `last_heartbeat_at` to `null`.

### 6. Clear stale configuration delivery and acknowledgement state

Provisioning currently creates MQTT credentials but does not automatically publish `apply_config`. A replacement client therefore has not acknowledged the previous configuration even if the old read model says “clean”. Clear:

- `last_config_sent_at`;
- `last_config_sent_hash`;
- `last_config_ack_at`;
- `last_config_ack_hash`.

After re-provisioning, the bank should remain configuration-dirty until a fresh `apply_config` is sent and acknowledged by the new client.

### 7. Do not log provisioning tokens or token-bearing topics

The token appears in the registration topic, so removing only the explicit token log is insufficient. Redact/remove it from:

- `RegistrationHandler` lookup and invalid-token logs;
- shared inbound-handler topic logging for registration messages;
- locker-client registration-topic logging.

Do not log even a shortened token.

### 8. Do not expose internal exception messages

Keep `$e->getMessage()` in structured server logs, but show a generic translated error in the Filament notification.

### 9. Strengthen tests

`test_reset_allows_a_previously_provisioned_bank_to_provision_again` currently only checks that `provisioned_at` is null and does not attempt a second provisioning. Extend coverage for:

- actual re-provisioning and `LockerWasProvisioned`;
- service authorization;
- actor/timestamp audit data and absence of plaintext token data;
- MQTT-user deletion and auth/ACL denial while deleted;
- old/new provisioning token behavior;
- offline and config-state cleanup;
- repeated resets and failure handling;
- log assertions for tokens and token-bearing topics.

### 10. Proposed ADR added

The agreed reset boundary is documented in `docs/adr/0029-administrative-locker-provisioning-reset.md`. It records this PR as an administrative recovery workflow, keeps tokens out of events, defines service authorization and state cleanup, and explicitly defers HMAC storage to #159 and durable MQTT credential identities to #161.

Keep the ADR status `Proposed` while the implementation is updated and reviewed; change it to `Accepted` only when the PR conforms to the documented decision.

The unrelated mobile `.gitignore` hardening may remain in this PR as agreed during review.

Tracking details were also updated in #110.

merona-merlin (changes requested):
Requesting changes based on the agreed provisioning-reset scope.

Please implement the requirements documented in the detailed review comment and ADR-0029 before merge. The key blockers are keeping the token out of the event store, service-level authorization, actor-aware auditing, MQTT-user deletion, offline/config-state cleanup, safe logging/error handling, and complete workflow tests.

HMAC storage and durable per-provisioning MQTT identities are intentionally deferred to #159 and #161.

Aibolit66:
Review findings addressed in a0cf1050, on top of a merge of current `dev`.

### 1–3. Token out of the event store, service authorization, actor auditing

New `LockerProvisioningResetService` owns the reset. It enforces `lockerbank.configure` server-side — previously action visibility was the only thing guarding a credential rotation — then rotates the token, deletes the MQTT user and records the event in one transaction, with a `lockForUpdate` on the bank so two concurrent resets cannot both mint a token and leave the loser holding one that never works. The Filament resource only calls the service.

`LockerProvisioningReset` now carries `lockerBankUuid`, `actorUserId` and `resetAtIso8601`, and nothing else. It is registered in `AuditEventPresenter` under Devices & Lockers, rendering the actor and the bank with no token material.

### 4–6. MQTT user and state cleanup

The user is deleted via `MqttUserService::deleteUser()`. The projector then sets the bank offline with `connection_status_changed_at`, clears `last_heartbeat_at`, and wipes all four config sent/ack fields so the bank reads dirty until a replacement client acknowledges a fresh `apply_config`. No application-level unprovisioned guard was added, per the agreed boundary.

### 7–8. Logs and error surface

Redaction is centralised in `MqttTopicRedactor` (backend) and `redactTopic` (client) rather than remembered per call site, and nothing of the token survives — not even a prefix. Applied to `RegistrationHandler`, the three `AbstractInboundMqttHandler` sites, the three `InboundMqttProtocolGuard` sites, and the client's transport publish log.

**One addition beyond the listed sites:** `MosquittoAuthController` logged the raw topic on every ACL check, and Mosquitto ACL-checks every registration publish — so the token was hitting the log four times per attempt. That was the noisiest leak of all and is now redacted too.

Filament shows a generic translated message; `$e->getMessage()` stays in the structured server log.

### 9. Tests

`LockerBankProvisioningResetTest` covers real re-provisioning through `RegistrationHandler`, service-level authorization denial, actor/timestamp audit data with assertions that neither the old nor the new token appears anywhere in the stored event, MQTT-user deletion with auth (`200 allow:false`) and ACL (`403`) denial, old-token-dead/new-token-works, offline and config-state cleanup, repeated resets, transactional rollback on failure, and `Log::listen` assertions that no registration or ACL log line contains the token.

Pint, PHPStan and the full suite pass — 267 backend tests, 108 locker-client tests.

### Verified against a live broker

Run with the fleet simulator, not only in tests:

| Step | Result |
|---|---|
| Simulator connected as the existing device | online, heartbeating |
| Reset | `provisioned_at=NULL, offline, heartbeat=NULL, config sent/ack=NULL, mqtt_users=0` |
| Old session after reset | still connected, but every publish refused — `ACL Default: Denied` |
| Cleared client identity, new token | re-provisioned; UUID username recreated with a fresh password |
| Token occurrences in the logs | 0 — all `locker/register/[redacted]` |

That run is what exposed the ACL logging leak.

### Also

- Reset added to the locker bank edit page, sharing one action definition with the list row so the two cannot drift in wording or authorization.
- German translations for the new strings.

### Open for you to decide

- **ADR-0029 is still `Proposed`.** The implementation now conforms, but flipping it to `Accepted` is a re-review judgement, not mine to self-certify.
- **The ADR number collides.** Current `dev` already carries `0029-admin-panel-locale-single-panel-session.md` and `0029-enforce-modbus-rtu-inter-frame-delay.md`, so there would be three 0029s on merge. Renumbering would move the file away from the path cited in the review, so I left it — say which you prefer.
- HMAC storage (#159) and per-provisioning MQTT identities (#161) remain out of scope, as agreed.

merona-merlin:
The earlier requested changes are now addressed on the current branch.

- `cc354db` replaces replayable plaintext storage with one-time HMAC tokens and a generation guard, implementing #159 alongside the reset workflow.
- `913479c` derives the HMAC subkey from validated `APP_KEY` via domain-separated HKDF, finalizes the one-time Filament flow, translations, and snapshot handling, and accepts ADR-0034.
- `73e7534` merges current `dev`; the conflicts preserve `dev`'s durable MQTT command-response recovery behavior and both translation sets.

Legacy open plaintext tokens are invalidated by migration and must be reissued. Durable MQTT identity/session revocation remains explicitly scoped to #161. There are no unresolved review threads; the existing `CHANGES_REQUESTED` review now needs a fresh reviewer decision.

merona-merlin (approved):

Issue #122, opened 2026-06-29
backend(filament): show user group memberships with link to group edit


merona-merlin:
## Context
Follow-up to PR #119 / #46 (ADR-0020). Groups are manageable via **Groups** resource, but admins cannot see from a **User** profile which groups a user belongs to.

## Goal
On **Users → Edit user**, add a relation manager (or equivalent) listing the user's active group memberships. Each row should link to the corresponding **Group** edit page (`GroupResource`).

## Acceptance criteria
- [ ] User edit page shows active group memberships (name, optional expiry)
- [ ] Clicking a group opens Group edit (members + compartment access tabs)
- [ ] Revoked/expired memberships handled consistently with ADR-0020 (hide or mark clearly)
- [ ] Admin-only, consistent with existing Group/User resources

## Notes
- Out of scope for PR #119; post-merge UX improvement.
- Suggested owner: Boris.

merona-merlin:
Implemented by merged PR #131: https://github.com/Open-Locker/Open-Locker/pull/131. The User edit page now includes a read-only Group memberships relation manager showing active membership name and expiry, linking each row to Group edit; revoked/expired memberships are filtered through `activeGroups()`. The PR checks passed: https://github.com/Open-Locker/Open-Locker/actions/runs/29151144721, and the current main backend quality run is green: https://github.com/Open-Locker/Open-Locker/actions/runs/32571557077.

PR #131, opened 2026-06-29
feat(admin): show user group memberships on User edit page (#122)

Aibolit66:


CrazyCowLady (approved):
I noticed some possible quality of life improvements, but they're only tangentially related, so I put them into chat. Display of group status in user edit page and linking to those groups works fine.
I'll also make a ticket to add a bit more testing, but that also doesn't have to happen in this PR.

PR #184, opened 2026-07-25
feat(locker-client): contract-aligned locker fleet simulator (#82)

### LINKED ISSUE #82: Build MQTT locker simulator for development
## Context

The current backend `mqtt:client-simulator` helped during early MQTT development, but it is drifting from the canonical contract and is becoming too limited for day-to-day development.

A reliable simulator is important for developing and testing backend, MQTT, mobile/API visibility, and locker-client workflows without physical hardware.

## Goal

Provide a contract-aligned MQTT locker simulator that can emulate a provisioned locker and common runtime behavior.

## Scope

Explore and implement the best simulator shape. Possible approaches:

- replace/refactor the backend `mqtt:client-simulator` command
- add a simulator mode to `locker-client`
- or provide a small dedicated simulator entrypoint that reuses locker-client MQTT helpers

The simulator should follow the AsyncAPI MQTT contract and avoid legacy payloads.

## Expected behavior

The simulator should support at least:

- connect with configured/provisioned MQTT credentials
- subscribe to `locker/{uuid}/command`
- respond to `open_compartment` with contract-valid command responses
- respond to `apply_config` with `applied_config_hash`
- publish heartbeats on `locker/{uuid}/state/heartbeat`
- publish retained compartment snapshots on `locker/{uuid}/state/compartments`
- optionally simulate connection/LWT behavior on `locker/{uuid}/state/connection`
- allow simple scripted/manual changes to compartment `door_state`

## Contract requirements

- Use `message_id` and top-level `timestamp` on all MQTT messages.
- Use `transaction_id` on command responses.
- Use `compartment_number`, not backend-only IDs.
- Do not publish legacy `locker/{uuid}/status` or multiplexed `locker/{uuid}/state` payloads.
- Keep payloads valid against `docs/asyncapi/schemas`.

## Testing

Add tests for the simulator behavior where practical:

- generated response payloads match JSON Schemas
- heartbeats and snapshots use the split state topics
- duplicate command handling remains idempotent

## Open decisions

- Should the simulator live as a backend Artisan command, a locker-client simulator mode, or both?
- Should it support provisioning flow simulation, or start only from already provisioned credentials?
- Should simulation scenarios be configured via CLI flags, YAML, or interactive commands?

## References

- #39
- #77
- `docs/asyncapi/mqtt.yaml`
- `docs/mqtt_integration_plan.md`

Aibolit66:
Closes #82. Implements [ADR-0027](../blob/feat/82-locker-fleet-simulator/docs/adr/0027-contract-aligned-locker-fleet-simulator.md).

## What

Emulate one or many locker banks with no hardware — `pnpm sim`. The backend
cannot tell a simulated bank from a real one, so backend, admin panel, API and
mobile workflows become testable end to end.

**→ [docs/simulator.md](../blob/feat/82-locker-fleet-simulator/docs/simulator.md)** is the how-to.

## Shape

The simulator is a **second adapter behind the existing `LockerBusPort`, plus its
own composition root** — deliberately *not* a `--mode` flag through the production
boot path:

- `InMemoryLockerBus` is a sibling of `WaveshareModbusBusActor`; both implement
  the same port.
- Everything above the port — use cases, dispatcher, `outbound-envelope`,
  `mqtt-schemas`, dedup — is reused **unmodified**. The simulator cannot emit a
  payload the real client wouldn't, which is the anti-drift property #82 asks for.
- `src/main.ts` and `src/bootstrap/createApp.ts` are untouched. The production
  diff is empty.

Fleet mode is one process, N banks, each with its own bus, dedup store,
credentials and connection. A single bank is a fleet of one, so there is no
separate single-device path.

## Two constraints found by running it

1. **A bank provisions exactly once.** `LockerBankAggregate::provision()` refuses
   once `provisioned_at` is set, and the backend never stores MQTT passwords in
   plaintext — so the credentials issued on the first run are the only ones. They
   are cached beside the scenario file (git-ignored, `0600`, never under
   `DATA_DIR`). Without this the simulator would be single-use per bank.
2. **No `state/connection` "online" message.** Only `state-connection-lost.json`
   exists and its `status` is `const: "offline"`. Publishing "online" would break
   contract conformance, and the real client doesn't either — LWT only. ADR-0027
   records this; adding one is a contract change for its own ADR.

## Also here

- Retires the backend `mqtt:client-simulator` command. The simulator covers what
  it did plus the compartment snapshots and connection LWT it never published;
  nothing referenced it (no test, CI job, Justfile target or compose service) and
  it still emitted a legacy `type: command_response` field.
- Interactive console (`list`, `open|close|unknown <bank> <n>`) and MQTT traffic
  logging on by default — watching the traffic is the point of a simulator.
- Refuses to start when `APP_ENV`/`NODE_ENV` is production unless overridden.

## Verification

Live against the local broker and backend, not only unit tests:

- both banks provisioned, connected, heartbeating
- seeded door states landed in the read model
- **full round trip**: `LockerService::openCompartment()` → MQTT → simulator →
  fake relay → retained snapshot → projector → compartment shows `open`
- mobile app reflected the change live over Reverb

`pnpm test` 153 passed · `pnpm check` clean · backend `composer analyse` clean,
256 backend tests pass after removing the artisan command.

## Notes for review

- ADR-0027 is still **Proposed** — flip to Accepted if the shape looks right.
- The simulator reads Filament-internal Alpine state nowhere; it does read
  `groupVisibility`-free production code only. It does depend on
  `paths.ts`-derived stores being process-global, which is why the simulator uses
  in-memory stores rather than refactoring production ones (Alternative E in the ADR).
- Unrelated backend gap noticed while testing, **not fixed here**: a freshly
  provisioned bank stays `connection_status = unknown` while heartbeating
  normally, because `LockerHeartbeatHandler` only emits `LockerConnectionRestored`
  when the bank was already `offline`. Documented in the troubleshooting section;
  worth its own issue.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `feat/167-compartment-list-navigation` only — 24 files, +2,660/−400.

### What's good, and verified

The composition root reuses the production path rather than reimplementing it — real `CommandDispatcher`, real handlers, real use cases, envelope, dedup and schemas, with only the Modbus bus and the file-backed stores swapped. That is what makes payloads contract-valid by construction, and it checks out from the imports rather than only from the comment.

Test coverage maps directly onto issue #82's contract requirements: `message_id` + top-level `timestamp` on every message, `transaction_id` idempotency (repeat → no second response, no second open), retained compartment snapshots, heartbeat published *not* retained, `apply_config` answering with the applied hash, an unmapped compartment returning a contract-valid error, and an explicit "legacy topics are never published" test.

The traffic logger was also checked for secret leakage: `summarize()` is allowlist-based, extracting only `compartments`, `uptime_seconds`, `action`/`result`/`error_code`/`transaction_id`, never dumping a whole payload. Provisioning replies cannot leak a password through it. Not an issue.

### 1. The credential cache can land outside the ignore rules — *still live at tip*

`simulator-scenario.yml` and `.simulator-credentials.json` are ignored — but only in `locker-client/.gitignore`. The cache is written *beside the scenario file*, whose path the operator chooses (`SimulatorCredentialCache.forScenario`). Point a scenario at the repo root, `docs/`, or anywhere outside `locker-client/`, and the resulting file — holding a live MQTT password — is no longer ignored and can be committed. The repo root `.gitignore` has no simulator entries.

Either add both patterns at the root, or pin the cache to one fixed directory instead of following the scenario file.

### 2. A corrupt cache bricks the simulated bank — *still live at tip*

`read()` swallows parse errors and returns `{}`, commented as "a corrupt cache should cost a re-provision attempt, not a crash." But per this class's own docblock, `LockerBankAggregate::provision()` refuses once `provisioned_at` is set — so that re-provision attempt is *always* rejected with "Locker bank is already provisioned."

The practical result: one truncated write and that bank is unusable until an admin resets its provisioning token. The graceful degradation does not degrade to anything. A loud failure naming the recovery path would serve better than a silent reset.

### 3. `mode: 0o600` only applies on creation — *still live at tip*

`fs.writeFileSync(..., { mode: 0o600 })` sets permissions when the file is created and is ignored when it already exists. If the cache was ever created under a looser umask, later writes silently keep the loose mode. Worth an explicit `chmod` for a file holding a live password.

### 4. ADR-0027 is still `Status: Proposed` — *still live at tip (renumbered to ADR-0031, status unchanged)*

The implementation lands here, so it should move to `Accepted` on merge. The ADR was rewritten in place and its date moved `2026-06-24` → `2026-07-25`; that is fine under the project rule since it was `Proposed`, not `Accepted` — no supersede needed. Just the status.

### 5. Stale reference to the deleted command — *still live at tip*

`TODO.md:35` still cites E2E tests "via `mqtt:client-simulator`/Artisan" after `MqttClientSimulatorCommand.php` was removed. The ADR's references are deliberate — it documents the retirement — but `TODO.md` is not.

### 6. Design note — 391-line composition root mirroring a 203-line one — *still live at tip, now 447 vs 241*

`createSimulatorApp.ts` is nearly twice the length of the `createApp.ts` it "deliberately mirrors step for step." The reuse argument is sound, but the duplication is real and will need re-syncing by hand whenever the production root changes. Worth deciding whether the shared wiring gets extracted before a third caller appears.

### Verdict

**Approve with follow-ups.** Items 1–3 are all in the credential cache and worth fixing together — item 1 is the one to insist on, since it can put a live MQTT password into git history. Items 4 and 5 are one-liners.

Aibolit66:
## Independent review by AI

**Scope:** issue #82 / PR #184 only. Delta plus the follow-up work: commit `fd8d4454` and the round that follows it on `chore/stack-review-followups`. Verified against tip content; static review, so the 289 / 288 pass / 1 skipped figure is the author's.

### Findings

| # | Finding | Status |
|---|---|---|
| 1 | Credential cache could land outside the ignore rules | **Fixed** — `fd8d4454` |
| 2 | Corrupt cache resets silently, re-provision then refused | **Declined — agreed**, reasoning needs correcting |
| 3 | `mode: 0o600` applies on creation only | **Declined — agreed**, reasoning needs correcting |
| 4 | ADR still `Proposed` | **Fixed** — ADR-0031 `Accepted` |
| 5 | Stale `TODO.md:35` reference to the deleted command | **Fixed** |
| 6 | Composition root duplication (477 vs 321) | **Addressed** — parity test, not extraction |
| 7 | Ignore fix partial: renamed scenario, env override | **Fixed** — documented |
| 8 | Production guard read one variable, not both | **Fixed** |

### The one real defect

Item 8. `(process.env.APP_ENV ?? process.env.NODE_ENV ?? '')` meant that once `APP_ENV` was set to anything — including an empty string, which is not nullish — `NODE_ENV` was never read. `APP_ENV=local NODE_ENV=production` started the simulator against production, publishing fake state under a real locker UUID, which is precisely what the guard's own docblock says would corrupt live read models. Not exotic: a Laravel-shaped `.env` beside a Node runtime is this repo's layout.

The fix collects both, filters empties, and searches for a production value. The `filter` is the load-bearing part — it is what stops an empty `APP_ENV` masking the real one, which coalescing alone could not. Extracted to `src/adapters/simulator/production-guard.ts` because `main-simulator.ts` ends in a bare `main().catch(...)` at module scope, so importing it to test the guard would boot the simulator. That is a real reason the function was untestable in place, not a convenience. Seven tests; exactly two fail against the old line.

### Item 6, and why the test is the right answer

Not extracting was correct — the simulator's value is being the production path with two substitutions, and shared wiring behind a common signature is how the two would diverge silently. The parity test pins the one property that must hold: both roots register the same handler set. Extracting independently, both yield `createOpenCompartmentHandler` and `createApplyConfigHandler`.

Two things make it a genuine alarm rather than a box tick: `assert.ok(production.length > 0)` stops the regex rotting into two empty arrays comparing equal, and the docblock names the divergences that are *deliberate* — no `recoverInterruptedCommands`, no flush wiring, because the dedup store is in-memory — so nobody later "fixes" it by widening it into a false failure.

Worth recording as a scope boundary rather than a defect: with an in-memory dedup store and no flush wiring, the simulator cannot exercise crash recovery or pending-response replay. Fine, unless someone reads "backend workflows testable end to end" as covering those paths.

### Credit

The secret-handling discipline held up under two separate probes. `summarize()` in the traffic logger is allowlist-based, so a provisioning reply cannot leak a password through it; and every scenario validation error names `bank.name`, never `provisioning_token` — including the duplicate-token check, where quoting the token would have been the natural phrasing. Consistent enough across unrelated code to read as deliberate.

### What is left

**Optional code change.** The parity test resolves sources through `process.cwd()`, so it works under `pnpm test` and fails with an unexplained ENOENT if `node --test` is run from the repo root. A sentence in the comment, or resolving upward to the package root, closes it.

**Two description corrections, needed for the declines to stand on the record:**

- Item 2 is fleet-wide, not per-bank. One file keyed by provisioning token; `read()` returns `{}` on corruption and the next `set()` writes that state back, so every bank in the scenario loses its credentials permanently. Recovery is an admin reset *plus* a fresh 64-char token pasted into the scenario file, per bank. Still not worth touching provisioning code — just a bigger paper cut than "confusing failure".
- Item 3 should say this class is normally the file's only writer, rather than that the path cannot produce a loose mode. `SIMULATOR_CREDENTIALS_FILE` can point at a file the operator created, and then the mode is inherited.

### Verdict

**Approve.** Items 1, 4, 5, 7 and 8 are verified fixed at tip; 2 and 3 are correctly declined; 6 is addressed in the form I would have chosen. The follow-up commits are tightly scoped — `fd8d4454` touches three files, all docs and config, and shadows nothing tracked — so the pattern seen on #218, where each fix introduced the next defect, did not repeat here.

The finding worth carrying forward: item 8 was a safety guard that had never been exercised, promising in `--help` what its code did not do. Both of the substantive items on this PR were guarantees asserted in prose with nothing behind them.

Issue #139, opened 2026-07-06
backend: atomic event-sourced write paths + uniform Filament error UX


Aibolit66:
## Context
Follow-up deferred from #128 / ADR-0028 (synchronous projectors, queued reactors).

With projectors now synchronous and `catch_exceptions=false` ("fail loud"), a
projector that throws mid-write leaves an **orphaned stored event** (event
persisted, read model not updated), recoverable only via
`event-sourcing:replay`. There is currently no transaction around `persist()`.

Note (verified in #128): Spatie's own `AggregateRoot::persistInTransaction()`
wraps only the event write and runs handlers **after** commit, so it does NOT
make event + projection atomic.

## Scope
1. **Atomicity** — wrap event-sourced write paths so the event + synchronous
   projection commit/roll back together (`DB::transaction(fn () => $agg->persist())`
   or equivalent).
2. **Reactor ordering under transactions** — queued reactors dispatched inside a
   transaction must use `after_commit` so they never run before the data is
   committed. This is a global queue/dispatch concern — evaluate blast radius.
3. **Reactor-level idempotency** — `DoorDetectionReactor` and `CommandResponseReactor`
   keep derivations idempotent with `derivedEventExists()` followed by `event()`.
   That pair is not atomic, so two consumers of the `events` queue handling one
   redelivery would each find nothing recorded and each emit. Duplicated events
   are permanent history and every replay counts the outcome twice. Currently
   safe only because both compose files run a single worker, recorded in
   ADR-0040's Risks. Exits: an advisory lock around check-and-emit (as
   `LastAdminGuard` already does), or a unique index over the transaction id per
   event class. Raised in review of #185.
4. **Uniform Filament error UX** — a consistent friendly error notification for
   failed event-sourced admin actions (several actions catch individually today;
   the Group/access actions rely on Filament's generic notification).

## Discussion / decisions needed
- Where to own the transaction boundary (service layer vs a shared helper).
- Whether `after_commit` should be global or scoped to event-sourcing dispatch.
- Whether orphaned-event detection/monitoring is worth adding.

## References
- #128, ADR-0028 (`docs/adr/0028-synchronous-projectors-queued-reactors.md`)
- `config/event-sourcing.php` (`catch_exceptions`, `queue`)

merona-merlin:
The beta scope is implemented in #231:

- event rows and synchronous projections share one transaction;
- Spatie's queued handler job is dispatched only after commit.

[...]

PR #185, opened 2026-07-26
feat: separate command acknowledgement from door-open detection (#94)

### LINKED ISSUE #94: Separate command acknowledgement from door-state open detection and alerts
## Context

The locker client currently sends an `open_compartment` command response with `result: success` after the unlock pulse was sent and initial relay monitoring completed. The backend currently derives this as `CompartmentOpened` and marks the open request as `opened`.

That conflates command execution with the physical door state. A successful unlock pulse does not guarantee that the compartment door actually opened.

This distinction becomes important for operational monitoring and future alerting: if a compartment door opens without a matching command, the system should be able to treat that as a potential break-in, tampering, faulty lock, or sensor issue.

## Scope

- Separate command acknowledgement from physical door-open detection.
- Treat `open_compartment` command success as "unlock pulse sent" / command execution acknowledgement, not physical `opened`.
- Detect physical opening from the configured door input / door-state signal.
- Correlate door-open detection with active or recent open commands.
- Mark an open request as physically opened only when a matching door-open signal is observed.
- Mark or surface a failure such as `OPEN_NOT_DETECTED` when no matching door-open signal is observed within the configured timeout.
- Support future alerting for door-open events without a matching active/recent command.

## Backend event-sourcing requirement

Backend domain behavior must remain event sourced. Any new backend domain event, status transition, alert, or correlation result must be represented as stored domain events and then projected into read models. Do not update domain state only through direct model writes outside the event-sourcing flow.

## Proposed direction

- Keep the client responsible for local hardware-level detection where practical, because it has the command context, compartment mapping, and sensor access.
- Keep `state/compartments` as the source of truth for current physical door state.
- Add explicit command outcome / door-open detection semantics rather than sending a second duplicate `command_response` for the same transaction.
- Update MQTT / AsyncAPI contracts and create or update an ADR if the external message contract changes.

## Open questions

- Should the client emit a dedicated `compartment_open_detected` / `compartment_open_failed` event, or should backend correlate existing door-state snapshots with pending commands?
- What timeout should be used for `OPEN_NOT_DETECTED`?
- How long should the backend consider a door-open event correlated with a previous command?
- What alert severity and workflow should be used for unauthorized door-open detection?

Aibolit66:
Closes #94. Implements [ADR-0031](../blob/feat/94-separate-ack-from-door-detection/docs/adr/0031-separate-command-acknowledgement-from-door-open-detection.md).

## The problem

The client reported `open_compartment` success once the unlock pulse was sent, and the backend recorded that as `opened`. A pulse is not an opened door: the lock can release while the door stays shut — jammed, blocked, held, worn latch.

There are four combinations of "relay fired" and "door opened". The system recorded one of them, and recorded it wrongly in half the cases.

| Relay fired | Door opened | Meaning | Recorded before |
| --- | --- | --- | --- |
| yes | yes | normal open | `opened` |
| yes | no | jam, blocked, bad latch | `opened` ❌ |
| no | yes | uncommanded open | not recorded ❌ |
| no | no | idle | n/a |

Both signals already existed and the client already read both — the relay coil and the Modbus door sensors, polled every 500ms. They were never connected to the open request that caused them.

## The shape

**The client decides.** Detection lives in `OpenCompartmentUseCase`: read the door before firing, watch it after, report `opened` / `already_open` / `door_jammed`. `execute()` still returns as soon as the pulse is sent, so acknowledgement stays immediate and detection runs behind it.

The backend deliberately does *not* correlate snapshot deltas with pending commands. Snapshots are change-only (ADR-0030), so an already-open door publishes nothing and a correlator would report failure for a compartment that is physically open.

**Outcomes travel on the existing event channel**, not as a second `command_response` — that would break inbound dedup (ADR-0002).

**Correlation anchors on the last relay fire**, unbounded, with the delta recorded. The relay fire rather than the command: a command that errored before reaching the lock cannot explain a door opening. No threshold is baked into the event stream; thresholds live in alerting policy where they can change without rewriting history.

`readDoorSensors` was already on `LockerBusPort`, so the client change is application-layer only — no new bus method, no new adapter, `adapters/modbus/` and `main.ts` untouched.

## Backend

`CommandResponseReactor` deriving `CompartmentOpenAcknowledged` instead of `CompartmentOpened` is the entire behavioural fix. The rest is the vocabulary it needs: five stored events, a `DoorDetectionReactor` that translates device events into domain events (mirroring how `CommandResponseReactor` translates generic responses), a status enum replacing the bare string, and projector cases.

| Status | Meaning |
| --- | --- |
| `acknowledged` | Unlock pulse sent. **Was `opened`.** |
| `opened` | Door observed open. |
| `already_open` | Door was open before the pulse. |
| `door_jammed` | Door never opened within the window. |

Legacy `CompartmentOpened` handlers are kept so replay of pre-ADR-0031 events still projects. Rows written before this change carry the old meaning and are not reinterpreted.

## Alerting

Every deviation is surfaced the same three ways: a red badge with a warning icon on the compartment, a danger toast broadcast over Reverb, and mail to holders of `compartment.open`. A successful open alerts nobody.

A jam is maintenance and an uncommanded open is security, but both mean somebody has to walk to that locker, so neither is made quieter than the other.

The badge is necessary because **a jammed compartment reports `door_state: closed` exactly like a healthy one** — the fault is in the lock, not the door position, so it cannot be expressed as a door state. It clears when a later open succeeds, i.e. when the lock is demonstrably working rather than when someone asserts it is.

## A bug only end-to-end testing found

Acknowledgement and detection are derived by two independent queued reactors. A door that opened in 501ms was projected *before* the acknowledgement that preceded it, and the late acknowledgement downgraded `opened` back to `acknowledged`. The projector now always records the acknowledgement timestamp but only moves the status forward. Two regression tests cover it. Unit tests could not see this — they deliver events in order.

## Simulator

Jam mode: a compartment that pulses the relay normally while its door stays shut. Without it the simulator could not produce the outcome this work exists to detect — `flashRelay` unconditionally opened the door. Seedable via `jammed:` in the scenario, togglable with `jam`/`unjam`.

Also fixes the console silently ignoring typed commands when stdin is not a TTY. It now says so and takes `--interactive` to force it on, which is how the alert paths could be verified from a script.

## Verification

Live against the broker and backend, all four paths:

| Case | Result |
| --- | --- |
| Jam | `acknowledged` → `door_jammed` exactly 10s later, badge, toast, 3 mails |
| Uncommanded open | recorded, toast, 3 mails |
| Already open | recorded immediately, toast, 3 mails |
| Healthy open | `opened`, `detection_ms: 500`, **silent** |

`composer quality` clean · 271 backend tests · `pnpm test` 170 client tests · contract tests assert the publisher's real output validates against the new schemas.

## Notes for review

- ADR-0031 is **Proposed** — flip to Accepted if the shape looks right.
- **Deploy needs a worker restart and `config:clear`.** Queue workers are long-running daemons holding old code, and a cached `bootstrap/cache/config.php` will not register the new reactors. Both bit me during testing.
- **Alert recipients** (`compartment.open` = Managers + Admins) are recorded in the ADR as the expected first thing to narrow if it proves noisy.
- **Detection timeout follows `heartbeat_interval_seconds`** rather than its own column. The coupling is a deliberate trade-off; a dedicated column is kept as Alternative D.
- **Mobile is unaffected** — it never subscribes to the open-status channel and never calls the open-requests endpoint, so its only feedback is the unchanged POST response. Telling a user their door did not open needs its own ticket.
- **Not done here:** `pnpm generate:api`. Regenerating surfaces ~300 lines of unrelated pre-existing drift and breaks typecheck in five files. This change adds two optional fields nothing reads. Worth its own housekeeping ticket.


merona-merlin:
@cursor review

Aibolit66:
## Independent review by AI

**Scope:** delta vs `feat/82-locker-fleet-simulator` only — 53 files, +2,956/−132.
Each finding is marked with whether it still holds at the stack tip (`docs/adr-number-collisions`).

### Verified good

Event-sourcing compliance is clean, which is what issue #94 explicitly demanded. `DoorDetectionReactor` only *reads* models (`LockerBank`, `Compartment`) to resolve identity and emits domain events; it never writes read-model state. The five new `StorableEvents` are registered in `config/event-sourcing.php`, projected by `CompartmentOpenRequestProjector`, with side effects isolated in reactors. The old conflation is handled honestly too — `CompartmentOpened` is kept with a "replay support only" note rather than being deleted out from under history.

Contract work is complete rather than implied: `docs/asyncapi/mqtt.yaml` updated, plus three payload schemas and four examples, and `mqttOutboundContract.test.ts` (+88) to hold it.

Test coverage is 14 cases in `DoorDetectionTest` covering the real branches — already-open as its own deviation, a jam recorded as failure not success, late acknowledgement not overwriting a physical outcome, uncommanded open alerting operators but *not* plain users, unknown compartment ignored, unrelated device events left alone.

### 1. Idempotency is check-then-act, so it races — *still live at tip*

`derivedEventExists()` followed by `event(...)` is not atomic, and the reactor is `ShouldQueue` on the `events` queue. Two workers handling a redelivered `DeviceEventReceived` for the same `transaction_id` can both pass the existence check and both emit — producing duplicate `CompartmentDoorOpenDetected` events, which is exactly what the guard exists to prevent.

`test_detection_derivation_is_idempotent` covers *sequential* redelivery, which passes either way; it cannot catch this. Safe only if the `events` queue runs a single worker. If that is the deployment assumption it should be written down; if not, this needs a uniqueness constraint or a lock rather than a read.

### 2. Two unindexed JSONB predicates per door event, on the hottest table — *still live at tip*

Every detection event runs `resolveOpeningRequest()` (`event_properties->commandId`) and `derivedEventExists()` (`event_properties->transactionId`) against `stored_events`. `event_class` is indexed, so each query is bounded to one class — but that class grows without bound: every `CompartmentOpeningRequested` ever recorded, forever, scanned on each door event. `event_properties` is `jsonb` with no GIN index.

It will not show up in testing and will degrade slowly in production. A GIN index, or a small correlation table keyed by transaction id, would bound it.

### 3. ADR numbers in code comments — *already fixed at tip*

This branch adds 12+ `ADR-0031` references in PHP comments plus two `ADR-0026`, against the project convention of keeping ADR numbers out of code and explaining the reason inline instead.

The renumbering is precisely why that convention exists: at the tip, `ADR-0031` is now the *fleet simulator* ADR, door detection became `0040`, and audit log became `0030` — so every one of those comments would now point at the wrong document. At the tip no ADR references remain in `locker-backend/app` or `locker-client/src` at all, so this is already resolved upstack.

### 4. ADR number collision introduced — *already fixed at tip by #214*

`0031-separate-command-acknowledgement-from-door-open-detection.md` lands alongside the existing `0031-website-url-path-localization.md`. (0030, 0032 and 0033 were already doubled before this branch.)

### Verdict

**Approve with follow-ups.** Items 3 and 4 need no action — later work resolved both. Items 1 and 2 are the substantive ones and both survive to the tip: item 1 needs a stated answer on `events`-queue worker concurrency, item 2 is a slow-burn index problem worth an issue rather than a blocker here.

Aibolit66:
## Independent review by AI

**Scope:** issue #94 / PR #185, across all review rounds. Verified at `ff0a4080`; static review, so test and PHPStan figures are the author's.

### Findings

| # | Finding | Status |
|---|---|---|
| 1 | Idempotency guard is check-then-act, so it races | **Resolved** — constraint documented, fix tracked as #139 item 3 |
| 2 | Two unindexed JSONB predicates per door event | **Deferred** — tracked as #222 |
| 3 | ADR numbers in code comments | **Fixed upstack** |
| 4 | ADR number collision | **Fixed upstack** |
| 5 | Forward-only rule guarded acknowledgement but not failure | **Fixed** — `ff0a4080` |
| 6 | Failure handler recorded a timestamp with no reason | **Fixed** — `ff0a4080` |

**1.** `derivedEventExists()` followed by `event()` is not atomic, and two consumers of the `events` queue would both emit on one redelivery — permanent history in an event store. Resolved by documenting the single-worker constraint rather than fixing it: the reactor docblock, both compose files, `CommandResponseReactor` (which shares the queue and the guard), and ADR-0040's Risks, which names both exits. The premise checks out — no `replicas` or `deploy:` in either compose file.

**2.** `event_properties->commandId` and `->transactionId` run against `stored_events` per door event, with `event_class` indexed but `event_properties` bare `jsonb`. Correctly deferred: measurement before a guessed index is right on an event store, and #222 records that reasoning.

**3 & 4.** Zero `ADR-00` references remain in `app/` or `src/`, and no duplicate ADR numbers — door detection is `0040`. Both resolved by #214.

**5.** Only `onCompartmentOpenAcknowledged` was ordering-guarded; the four outcome handlers wrote unconditionally. A command whose relay fired but whose response never published gets an error from restart recovery, which would flip a detected `opened` to `failed` — the exact conflation this PR exists to remove. Now `whereNotIn` covers `Opened`, `AlreadyOpen` and `DoorJammed`. The reverse direction stays unguarded, correctly: a physical outcome arriving after `Failed` should win.

**6.** The first version of that guard put `error_code` and `error_message` inside it, leaving a row reading `status = opened` with a populated `failed_at` and no reason — visible in two Filament tables and the API resource. Now all three are recorded unconditionally and only `status` is withheld, matching the acknowledgement handler.

### Verified good

Event-sourcing compliance is clean: `DoorDetectionReactor` reads models only to resolve identity and emits domain events, never writing read-model state. `CompartmentOpened` is kept with a replay-support note rather than deleted from under history. Contract work is complete rather than implied — AsyncAPI, three payload schemas, four examples, and a contract test holding them. `CommandResponseReactor` now guards both derivations symmetrically, closing the asymmetry noted on #218.

Test coverage reaches the branches that matter: already-open as its own deviation, a jam recorded as failure not success, late acknowledgement and late failure each unable to overwrite a physical outcome, and an ordinary failure still landing — that last one being the test that stops the guard being over-tightened.

### One loose end

`GET open-requests/{commandId}` can now return an `error_code` on a request whose `status` is `opened`. Verified as harmless today — `error_code` appears only in `src/store/generatedApi.ts`, no mobile code reads it, nothing calls the endpoint. But it is a contract semantic that isn't written down. A line in the endpoint's Scramble docs — *`error_code` records what the device reported and does not imply failure; read `status` for the outcome* — would save the next consumer a wrong guess.

### Verdict

**Approve.** All six findings are fixed, correctly deferred with a ticket, or resolved upstack. Both deferrals are tracked with their reasoning (#139 item 3, #222), which is what separates a decision from an omission.

Worth recording for whoever merges: item 6 existed only because a change summary described behaviour the code did not have. Comparing the description against the diff — not reading the diff alone — is what surfaced it, for the second time on this stack.

PR #189, opened 2026-07-28
feat: trace and log the open flow end to end with OpenTelemetry (#65)

### LINKED ISSUE #65: Add OpenTelemetry to backend and iot client (plus usage docs)
## Goal
Add OpenTelemetry-based observability across backend + IoT clients so we can trace an "open compartment" flow end-to-end, and document how to use it.

## Important constraint
Do **not** add SigNoz (or any heavy observability stack) to the default docker setup.

### 1) Define a project-wide OpenTelemetry strategy
- Traces first; logs/metrics optional
- Naming conventions for `service.name`
- Standard span attributes:
  - `command_id`, `transaction_id`, `message_id`, `locker_uuid`, `compartment_number`
- Decide how to propagate context across MQTT (where feasible)

#### Backend (Laravel/PHP)
- Instrument:
  - HTTP requests
  - queue jobs
  - mqtt-listener processing spans
- Ensure correlation between logs and traces where possible (trace_id/span_id in logs)

#### IoT client (Node locker-client)
- Instrument:
  - MQTT publish/subscribe spans
  - Modbus operation spans
- Add the same correlation attributes as the backend

### 3) Provide an optional collector setup (opt-in)
- Provide an **optional** OpenTelemetry Collector setup (local or external)
- Provide a minimal example configuration to export to a backend of choice (SigNoz, Grafana Tempo, Honeycomb, etc.)
- If we provide docker compose, it should be a separate opt-in overlay (e.g. `docker-compose.observability.yml`), not enabled by default

### 4) Documentation
- How to run locally (with/without collector)
- Required env vars
- How to trace one full open flow (API → MQTT command → client → Modbus → MQTT response/event → backend)

## References (examples)
- SigNoz Laravel OTel instrumentation: https://signoz.io/docs/instrumentation/opentelemetry-laravel
- SigNoz Node.js OTel instrumentation: https://signoz.io/docs/instrumentation/javascript/opentelemetry-nodejs/

Aibolit66:
Closes #65.

An open request crosses four processes, and until now each one logged independently. When a compartment did not open, a command that never reached the broker, a client that never answered, and a reactor stuck on the queue all looked identical from the API side. This adds distributed tracing and log shipping across the backend and the locker client, so the whole flow is one timeline.

> **Stacked on #185.** Base is `feat/94-separate-ack-from-door-detection`; review that one first.

## What's done

**Strategy — `docs/adr/0032`**
- Traces *and* logs, both components. Metrics out of scope.
- Hybrid instrumentation: auto for the framework edges (HTTP, queue, DB), hand-written for MQTT and Modbus.
- `keepsuit/laravel-opentelemetry` over the official SIG package, because the latter declares `ext-opentelemetry` — a native extension that would become mandatory in the image, in CI, and in every `composer install`.

**Backend**
- Producer span on every outbound publish, consumer span on every inbound message — both at the single choke point of their direction, so a new topic cannot skip instrumentation.
- HTTP, queue jobs, database and Redis come from the package; a queued reactor lands under the request that caused it.
- Log records ship over OTLP through an `otlp` channel *added to* the existing stack, never replacing it.
- Broker authorization decisions moved to their own `broker` channel, out of the telemetry stack.

**Locker client**
- `TracingPort` + `LogShippingPort` in `src/ports`, one OpenTelemetry adapter behind both, imported lazily — a Pi without a collector never loads the SDK.
- Spans for MQTT dispatch/publish and every Modbus operation. The Modbus span opens *before* the operation is queued, so time spent waiting behind other bus traffic is visible.
- Logs reach the collector via a winston transport added next to the console one, which stays.
- The fleet simulator is wired identically, so a full flow can be demonstrated without hardware.

**Contract**
- Optional `traceparent` on every MQTT envelope, declared once in `schemas/common/traceparent.json`. Every envelope already allowed additional properties, and the client's zod schemas strip unknown fields, so this is non-breaking in both directions.

**Operations**
- `docker-compose.observability.yml`: opt-in overlay adding an OTLP collector and pointing every PHP process at it. **No trace UI ships with it** — the collector forwards to whichever backend you run separately, keeping that choice out of the repo and the default stack untouched.
- `docs/observability.md`: the runbook — env vars, the startup sequence that actually works, what a healthy trace looks like, how to read one when a hop is missing, and where logs and exceptions appear.

## Volume control

Sampling stays at 100%; volume is controlled by not instrumenting the chatter. Heartbeats and retained snapshots produce no spans on **either** side, and broker-auth logs stay out of the dashboard. A 10% sample would most often discard the incident being investigated.

## Off by default

Nothing is recorded or exported until `OTEL_EXPORTER_OTLP_ENDPOINT` is set. With no endpoint the backend adds nothing to MQTT payloads and the client never loads the SDK. Turning it on is configuration, not a deploy.

## Four bugs found by running it

None of these were reachable from the test suite; each needed a live broker and a long-running process.

| Bug | Effect | Fix |
|---|---|---|
| `mqtt:listen` is not a worker mode the package detects | Its spans never flushed — the return half of every trace was invisible | Explicit flush per traced message |
| Shared log context outlived its span | Every later line, heartbeats included, carried a dead trace id — following it led to an unrelated incident | Context cleared in a `finally` |
| Client traced its own heartbeats | A span per heartbeat, plus `traceparent` on the chattiest topic | State topics excluded on publish |
| `flush_after_each_iteration` defaults to false | A failing queue job exported **nothing** — the failure you most want to see | Enabled, documented for non-overlay deploys |

The first three are held by regression tests that were verified to fail without the fix. The fourth is config, guarded by documentation only.

## Verification

- Backend: **283 tests**, Pint and PHPStan clean.
- Locker client: **188 tests**, typecheck/format/lint clean.
- End-to-end, against a live broker and a running client: one trace, 131 spans, spanning both services from `POST /api/compartments/{compartment}/open` through the MQTT hop to the client and back.
- Logs from both services confirmed in the dashboard, carrying trace ids.
- A deliberate broker outage produced red spans and a stack trace in the exceptions view.

## Known gaps

- **Modbus spans have never run on real hardware.** The simulator uses an in-memory bus, so they are unit-tested only — and that is the hop this work most exists to illuminate.
- **No automated test for client log shipping**, verified by hand.
- **Heartbeat log volume** is high (~90% of log lines in a quiet period) and deserves the same exclusion treatment as broker-auth logs.
- The mobile app is not instrumented, so traces start at the backend, not at the tap. Out of scope for #65.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `feat/94-separate-ack-from-door-detection` only — 51 files, +4,684/−60.
Findings are marked with whether they still hold at the stack tip.

### Verified good

The issue's hard constraint — *no heavy observability stack in the default docker setup* — is respected properly, not nominally. The collector lives in a separate `docker-compose.observability.yml`, every `.env.example` entry is commented out, and `createTracing()` returns `noopTracing` unless `OTEL_EXPORTER_OTLP_ENDPOINT` is set. Better still, the OTel SDK is imported **lazily**, so a Pi running without a collector never loads it and pays no startup, memory or batching cost.

The MQTT contract change is handled correctly. `traceparent` is optional across all six message schemas, every envelope is `additionalProperties: true`, and `MqttTraceContext::extract()` treats a malformed header as "start a fresh trace" rather than a rejection. It is explicitly never persisted, never part of deduplication, never used for business correlation — `message_id` and `transaction_id` keep those roles. ADR-0032 covers the strategy, which is what the project rules require for a cross-component decision like this.

Two details worth calling out because they are easy to get wrong and were not: registration topics carry the provisioning token in the path, and `spanDestination()` templates it out of both the span name and the attribute — a secret is not one refactor away from a span. And the chatty `/state/` topics (heartbeat, retained snapshots) are excluded from tracing on **both** sides, symmetrically, so retained messages never carry a stale `traceparent` that would graft new subscribers onto a long-dead trace.

The Dockerfile change is scoped to a single `RUN`, so `OTEL_SDK_DISABLED` does not leak into the runtime image.

### 1. `state.json` gains a `traceparent` field that will never be sent — *still live at tip*

`docs/asyncapi/schemas/messages/state.json` adds `traceparent`, but `tracesTopic()` returns false for every `/state/` topic and the backend declines the same two topics on receipt. By design, no state message will ever carry the field.

Harmless at runtime, but the contract now advertises something the implementation deliberately never produces — the next person to read the schema will reasonably expect it. Either drop it from `state.json`, or add a line saying it is reserved and intentionally unused there.

### 2. Nothing suppresses exporter diagnostics when the collector is unreachable — *still live at tip*

`OtelTracingAdapter` configures the OTLP exporters and a `BatchSpanProcessor` but never sets a `diag` logger or level, so export failures fall through to OpenTelemetry's default console diagnostics — the familiar `Export failure … retry limit exceeded … connect to localhost:4318`, repeating for as long as the endpoint stays down.

On a Raspberry Pi with a configured-but-unreachable collector — a flaky site link, a collector that got redeployed — that repeats indefinitely and lands in the device's logs. Worth a `diag` level or a backoff/circuit breaker so telemetry failure stays quieter than the thing it is observing.

### 3. Default sampling is 100% — *still live at tip*

`.env.example` documents `always_on` as the default with `traceidratio` as the opt-in alternative. That is the right default for developing and for tracing one open flow end to end, which is what the issue asked for. It is the wrong default for a fleet, and the moment someone sets an endpoint in production they get every span.

Not a code change — a deployment note in `docs/observability.md` saying "set a ratio before pointing this at production" would cover it.

### Verdict

**Approve.** All three items are documentation or configuration judgement calls, not defects — nothing here blocks merge. Item 1 is the one worth fixing while it is cheap, since a wrong contract outlives the person who wrote it.

Aibolit66:
## Independent review by AI

**Scope:** issue #65 / PR #189, across all review rounds. Verified against the current working tree; static review — findings come from reading the diff, the OpenTelemetry package source in `node_modules`, and the schemas, so test and lint figures are the author's.

### Findings

| # | Finding | Result |
|---|---|---|
| 1 | `state.json` advertises a `traceparent` that is never sent | **Fixed** — documented as reserved |
| 2 | Exporter diagnostics unhandled when the collector is unreachable | **Fixed** — throttled reporting added |
| 3 | Sampling defaults to 100% | **Fixed** — deployment note |
| 4 | Simulator got no logger, so its diagnostics were discarded | **Fixed** — both roots wired alike |
| 5 | Throttle key never matched for id-bearing messages; map unbounded | **Fixed** — normalisation + cap |
| 6 | `diag.setLogger` re-registered per simulated bank | **Fixed** — override message suppressed, constraint documented |

### 1 — a contract advertising what it never produces

`traceparent` was added to `state.json` while `tracesTopic()` returns false for every `/state/` topic on both sides, so no state message can ever carry it. Now carries a description saying it is reserved and why — heartbeats and retained snapshots would dominate the trace backend. Keeping the field rather than deleting it is the right call: the six envelopes stay symmetrical, and the next reader is told why this one is empty.

### 2 — reporting where there was silence

The original finding said an unreachable collector would spam the Pi's console. **That was wrong, and the correction matters for what this change is worth.** `DiagAPI`'s log proxy short-circuits when no logger is registered (`if (!logger) return`), and `DiagConsoleLogger` is opt-in and was never wired up — so exporter failures were being dropped silently, not printed repeatedly.

So this is not noise suppression; it is new reporting. `quietenExporterDiagnostics` routes the SDK's own diagnostics through the app logger, once per distinct message per five minutes, with the first occurrence always reported. That turns a silent failure mode — a collector that stops accepting spans while the client carries on as if traced — into one warning that says so. Worth having on those terms.

### 3 — a default that is right for one flow and wrong for a fleet

`always_on` suits development and tracing a single open end to end, which is what the issue asked for. `docs/observability.md` now says to set a ratio before pointing a deployment at a collector, "or accept the volume deliberately rather than by omission" — directive rather than descriptive, which is what makes a deployment note usable.

### 4 — the two composition roots disagreed

`createSimulatorApp` called `createTracing({ lockerUuid })` with no logger, falling through to `noopLogger`, so the component that exists to exercise the tracing path without hardware was the one place exporter failures went nowhere. Now passes `log: logger`, with the reason recorded inline. Both roots wired identically.

### 5 — the throttle that would not have throttled

The key was the raw message string, and the SDK embeds identifiers in some warnings — `Cannot execute the operation on ended Span {traceId: …, spanId: …}`, and the `end() called twice` message carrying `traceId-spanId`. Every occurrence is a distinct string: never throttled, and each one a permanent map entry on a device that runs for months.

Both halves fixed. `throttleKey` collapses `\b[0-9a-f]{16,32}\b`, covering 32-char trace ids and 16-char span ids including the hyphenated pair, and truncates at 200 characters so key size is bounded too. The map is capped at 256 with oldest-first eviction, and the constant's docblock is honest that the cap is a backstop for an unforeseen shape rather than the primary mechanism. A test drives 20 warnings with differing ids and asserts one report.

### 6 — a process-global hook configured per device

`createTracing` runs inside `startSimulatedDevice`, so a simulated fleet re-registers the global diag logger once per bank, each replacing the previous throttle window — and the SDK emitted an override warning each time, which the earlier test had to filter around. Now `suppressOverrideMessage: true`, verified as real in the installed `@opentelemetry/api@1.9.1` (`setLogger(logger, options?: DiagLoggerOptions)`, honoured at `diag.js:62`), with a docblock recording that the last registration wins.

### Verified good

The issue's hard constraint holds properly: the collector lives in a separate compose overlay, every `.env.example` entry is commented out, and the OTel SDK is imported lazily, so a Pi without a collector never loads it. Registration topics carry the provisioning token in the path and `spanDestination()` templates it out of both span name and attribute — a secret is not one refactor away from a span. `/state/` topics are excluded from tracing symmetrically on both sides, so retained messages never graft new subscribers onto a dead trace. `MqttTraceContext::extract()` treats a malformed header as "start a fresh trace" rather than a rejection.

### Two things left, neither worth acting on

`Map.set` on an existing key does not refresh insertion order, so eviction is first-seen rather than least-recently-used — reachable only in the scenario the cap exists to backstop. And `throttleKey` will not normalise dashed UUIDs or hex runs longer than 32 characters; neither is a shape the SDK currently emits.

### Verdict

**Approve.** Six findings, all closed. Items 1 and 3 were contract and documentation judgement calls; 2, 4, 5 and 6 turned a silent failure mode into a bounded, reported one, wired the same way in both composition roots.

Worth recording for whoever merges: item 5 existed only because the fix for item 2 introduced it, and item 2's original justification was itself wrong about the baseline behaviour. Neither would have surfaced without re-reading the SDK source rather than trusting the description of what it does.

PR #191, opened 2026-07-29
fix(admin): repair status badges and align the two compartment views (#190)

### LINKED ISSUE #190: Check filament bugs with simulator and new open door/relay fire separation

Aibolit66:
Fixes #190.

Found by running the fleet simulator against the panel after #94 separated command acknowledgement from door-open detection.

## The 500s

#94 cast `CompartmentOpenRequest::$status` to `CompartmentOpenRequestStatus`, but three Filament tables still type-hinted `?string $state`. Filament passed the enum, PHP threw a TypeError, and the page 500'd — on the locker bank edit page and the user edit page. `CompartmentOpenRequestResource` had already been moved to the enum pattern; the relation managers were missed.

Colour and label now come from the enum itself, so a badge cannot drift from the cast again, and the status filter is built from `cases()` instead of a hardcoded list that predated `acknowledged`, `already_open` and `door_jammed`.

## Consistency between the two compartment views

The locker bank's compartment table showed a "Last open status" column while the compartment list showed a door badge — two different shapes for the same subject. The bank page now uses the same door badge; the full command lifecycle is one tab away under Open command history. The badge and its fault markers moved into a shared `CompartmentDoorStateColumn` so the two screens cannot drift apart again.

## Open button

Hidden once the door is open — the command would be a no-op the device answers with `already_open`. It also now checks the `CompartmentOpen` permission, which the bank page never did.

## Stale fault marker

Only a jam marks a compartment faulty. `already_open` describes one past command, and nothing a healthy door does afterwards creates a newer open request — so a compartment stayed flagged red forever after a single already-open response, which is what surfaced during simulator testing.

## Also

- German translations for the nine status labels, which were missing entirely
- `just trace-up` / `trace-overlay` / `trace-status` / `trace-down`, after tracing silently broke: the stack had been started without the observability overlay, so the backend exported nothing while every container looked healthy. `trace-status` checks for exactly that case.

## Verification

Pint, PHPStan and the full suite (301 tests) pass. New `OpenRequestStatusBadgeTest` renders every status through all three tables. Verified end to end against the simulator and the mobile app: an open went through as `granted_access`, ack and door-open were recorded separately with a 501 ms detection, and it produced a single 84-span trace across backend and client.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `feat/65-opentelemetry-tracing` only — 11 files, +378/−81.

### Verified good

The core of this is a genuine de-duplication: `CompartmentDoorStateColumn` pulls the door badge, its fault marker and the "can this be opened" rule into one place, so the compartment list and the locker bank's compartment table can no longer disagree about the same door. The docblock is honest about why — they had already drifted once.

The reasoning encoded in it holds up. Only `DoorJammed` flags a fault, and the comment explains precisely why `AlreadyOpen` must not: nothing a working door does afterwards creates a newer open request, so flagging it would mark the compartment faulty forever. That is the kind of thing that looks like an oversight until you think it through.

`CompartmentOpenRequestStatus::color()` also gets the important call right — `Acknowledged` is *not* success. Now that the unlock pulse is separated from door detection, the pulse being sent says nothing about the door having moved, and colouring it green would undo that distinction in the UI.

The type change from `?string $state` to `?CompartmentOpenRequestStatus $state` is safe: `CompartmentOpenRequest` casts `status` to the enum. And the new `OpenRequestStatusBadgeTest` drives **every** enum case through both tables via data providers — exactly the test that would catch a cast or type mismatch.

### 1. Issue #190 has an empty body — *still live at tip*

No acceptance criteria, so the PR effectively defines its own scope and there is nothing independent to verify it against. Not a code problem, but it means "does this fix the reported bugs?" is not an answerable question from the ticket.

### 2. N+1 left behind in the one table this PR did not align — *still live at tip*

`CompartmentAccessesRelationManager` reads `compartment.latestOpenRequest.status` and `.opened_at` — two hops — with no eager loading, while this PR correctly adds `->with('latestOpenRequest')` to both sibling tables (`CompartmentResource:104,136`, `CompartmentsRelationManager:107`).

Pre-existing rather than introduced here, but this PR edits that exact column and fixes the pattern everywhere else, so it is the natural moment. A `modifyQueryUsing(fn ($q) => $q->with('compartment.latestOpenRequest'))` finishes the job.

### 3. Tracing tooling landed in the admin-bugs PR — *still live at tip*

`Justfile` (+107) adds `trace-overlay`, `trace-up`, `trace-down` and `trace-status` recipes, including SigNoz checkout and port management, plus `docs/observability.md` (+37). That is observability work — it belongs with the OpenTelemetry PR, not with "check filament bugs with simulator".

To be clear, it does **not** violate the constraint from #65: SigNoz arrives through an opt-in `just` recipe, not the default docker setup. The issue is reviewability — these two concerns now cannot be reviewed, reverted or merged independently.

### Verdict

**Approve.** Item 2 is a small, worthwhile addition while the file is open; item 3 is a note for how work gets split next time, not something to unpick now. Nothing here blocks merge.

Aibolit66:
## Independent review by AI

**Scope:** issue #190 / PR #191, across both review rounds. Verified against the current working tree; static review — findings come from reading the diff, the models and the Filament tables, so test and PHPStan figures are the author's.

### Findings

| # | Finding | Result |
|---|---|---|
| 1 | Issue #190 has an empty body | **No action** — noted for closing |
| 2 | N+1 in the one table this PR didn't align | **Fixed** — in both relation managers |
| 3 | Tracing tooling landed in the admin-bugs PR | **No action** — already merged upstack |

### 2 — the actionable one, and it went further than the finding

`CompartmentAccessesRelationManager` rendered `compartment.latestOpenRequest.status` and `.opened_at` with no eager loading, while the same PR added `->with('latestOpenRequest')` to both sibling tables. Fixed two places, left the third.

The fix covers more than was asked. Alongside `compartment.latestOpenRequest` it loads `compartment.lockerBank`, which the `compartment.lockerBank.name` column needs and which the original finding missed — and `grantedByUser` / `revokedByUser`, which are resolved inside `->state()` closures rather than dotted column names. Both of those columns are `toggleable()` without `isToggledHiddenByDefault`, so they render by default and were a query per row each.

The comment records why they were easy to miss: a dotted column name makes a relation hop visible, a closure hides it. That is the part worth keeping — the next person adding a `->state()` closure to this table has the warning in front of them.

`UserAccessesRelationManager` was taken too, adding `user` and `grantedByUser` to its existing `->active()` modifier. Omitting `revokedByUser` there is correct rather than an oversight: that table is filtered to active grants and has no revoked column.

Taking the adjacent table was the right call — leaving it would have repeated the exact pattern the finding was about.

### 1 — no acceptance criteria

Nothing to fix in code. It matters only at closing time: with an empty ticket body, "does this fix the reported bugs?" is not answerable from the ticket, so the PR defines its own scope. Worth a line on the issue when it is closed.

### 3 — reviewability, not a violation

The `Justfile` tracing recipes and the `docs/observability.md` growth are observability work in an admin-bugs PR. It does not breach #65's constraint — SigNoz arrives through an opt-in recipe, not the default docker setup — and it is already merged upstack, so there is nothing to unpick. The cost is that the two concerns can no longer be reviewed or reverted independently.

### Verified good

The de-duplication is real: `CompartmentDoorStateColumn` puts the door badge, its fault marker and the "can this be opened" rule in one place, so the two compartment views can no longer drift — which the docblock admits they already had. Only `DoorJammed` flags a fault, and the comment explains why `AlreadyOpen` must not: nothing a working door does afterwards creates a newer open request, so flagging it would mark the compartment faulty forever.

`CompartmentOpenRequestStatus::color()` gets the consequential call right — `Acknowledged` is not success. With the unlock pulse now separated from door detection, colouring it green would undo that distinction in the UI. `OpenRequestStatusBadgeTest` drives every enum case through both tables via data providers, which is the test that catches a cast or type mismatch.

### Noted for elsewhere

The same defect exists in four more admin tables that resolve a relation inside a `->state()` closure with no eager loading — `GroupResource:88`, `TermsDocumentVersionResource:72,82`, `UserResource:162,165` and `AuditLogResource:88`. `UserResource` is the sharpest: `latestAcceptedTermsVersion()` runs `termsAcceptances()->with('acceptedVersion')->latest()->first()` per row, where the inner `->with()` makes it look eager-loaded while the outer query is not. Out of scope here and tracked separately.

### Verdict

**Approve.** Item 2 is fixed in both relation managers and covers both hop styles; items 1 and 3 need no code change. The panel-wide sweep is tracked elsewhere, which is the right home for it — it is a Filament-wide pattern, not an admin-bugs fix.

Issue #173, opened 2026-07-14
Improve locker-client protocol errors and observability


merona-merlin:
## Context

Some invalid commands are silently dropped and hardware, polling, and publish failures do not yet provide consistent operational context.

## Scope

Clarify protocol error responses, add structured diagnostics and health signals, and preserve useful context without exposing credentials.

Aibolit66:
Some of the surrounding scope is already covered — schema-invalid commands *do* get an `INVALID_COMMAND` response, `MqttErrorCode` gives us a typed vocabulary, the Last Will publishes connection state, and logging is already structured JSON. What follows is what's actually still silent.

## Invalid commands that are silently dropped

All in [`adapters/mqtt/command-dispatcher.ts`](https://github.com/Open-Locker/Open-Locker/blob/9076687d/locker-client/src/adapters/mqtt/command-dispatcher.ts). Each logs a warning and returns without publishing a `command_response`, so the backend waits on a reply that never arrives.

| Case | Correlation id available | |
|---|---|---|
| Payload isn't valid JSON | none — nothing parsed | [L41](https://github.com/Open-Locker/Open-Locker/blob/9076687d/locker-client/src/adapters/mqtt/command-dispatcher.ts#L41) |
| `action` missing or not a string | `transaction_id` may be present | [L47](https://github.com/Open-Locker/Open-Locker/blob/9076687d/locker-client/src/adapters/mqtt/command-dispatcher.ts#L47) |
| `action` has no registered handler | `transaction_id` may be present | [L53](https://github.com/Open-Locker/Open-Locker/blob/9076687d/locker-client/src/adapters/mqtt/command-dispatcher.ts#L53) |
| Guard rejected: `missing_message_id` | `transaction_id` may be present | [L61](https://github.com/Open-Locker/Open-Locker/blob/9076687d/locker-client/src/adapters/mqtt/command-dispatcher.ts#L61) |
| Guard rejected: `missing_transaction_id` | none | [L61](https://github.com/Open-Locker/Open-Locker/blob/9076687d/locker-client/src/adapters/mqtt/command-dispatcher.ts#L61) |
| Guard rejected: `duplicate_message_id` | both present | [L61](https://github.com/Open-Locker/Open-Locker/blob/9076687d/locker-client/src/adapters/mqtt/command-dispatcher.ts#L61) |
[...]

PR #192, opened 2026-07-30
fix: improve locker-client protocol errors and observability (#173)

Aibolit66:
Covers #173 end to end: protocol error responses, structured diagnostics, and hardware health. Stacked on `fix/190-admin-bugs`.

## Invalid commands

Five of the eight paths that dropped a command now answer it:

| Case | Answer |
|---|---|
| `action` missing or not a string | `INVALID_COMMAND` |
| unregistered action | `UNKNOWN_ACTION` (new) |
| `missing_message_id` | `MISSING_MESSAGE_ID` (new) |
| `duplicate_message_id` | replays the stored response |
| `duplicate_completed` | replays the stored response |

Replay re-sends the answer a transaction already produced rather than an error. A retry usually means the backend never saw the first reply, so repeating it is what unblocks the transaction; an error would contradict a reply the backend may have acted on. The envelope is stamped with a fresh `message_id`, or the backend's own transport dedup would discard it. The `transaction_id` is unchanged, so a backend that did receive the original takes no second action.

Three stay silent on purpose:

- **Unparseable JSON** and **`missing_transaction_id`** carry no correlation id. Answering would mean synthesising one, which would resolve some unrelated pending command.
- **`duplicate_in_progress`** has no stored reply yet, and `result` is `enum: [success, error]` — there is no shape for work still running without changing the schema and the backend validator. The execution already under way publishes the real answer when it lands.

The two new error codes need no contract change: `error_code` is typed as a free-form string and the backend stores it opaquely.

## Hardware, polling, publish

`adapters/modbus/` contained no logging at all — failures were rethrown bare or converted to a fallback value, so an unreachable board was indistinguishable from closed doors. `LoggerPort` gains `error()` and is injected into the bus actor, RTU driver and reconnect coordinator. The bus actor owns reporting hardware failure; the polling use case keeps a logged backstop for exceptions that escape it, which resolves the double-swallow noted on the issue.

`MessageTransportPort.publish` now returns whether the broker accepted the message. It resolved either way before, so a command response lost while disconnected looked exactly like a delivered one. `OutboundMqttAdapter` reports a dropped message with its `message_id`, `transaction_id` and `action` — the transport only sees a topic and a string.

A boolean rather than a rejection, deliberately: the poll timer and inbound dispatch publish without awaiting, so a rejection there becomes an unhandled rejection that terminates the process. That fire-and-forget `dispatcher.dispatch` call had no `catch` and was already a latent crash path; it has one now.

## Error codes

`mapErrorToMqttCode` guessed codes by substring-matching messages (`includes('jammed')`, `'not found'`, `'modbus'`). Every path it matched already throws a typed `LockerError`, so the guessing could only mislabel a reworded message. It is now types first, with message matching confined to `isModbusLibraryError` for `modbus-serial` errors we do not author. `ModbusTransportError` gains a `reconnectable` flag so "port closed" is distinguishable from "library API missing" — both previously matched the same string check.

## Note on one changed assertion

`simulated-device.test.ts` asserted that a repeated `transaction_id` "must not publish anything further". ADR-0002 states business idempotency as *"do not execute one command twice"* — about side effects, not reply count — so replay is compatible, and the test encoded the old implementation rather than the contract. It now asserts one open, the same answer replayed, and a fresh `message_id`. Flagging it because it is the one place this branch contradicts an existing written assertion; if ADR-0002 is read as also forbidding a second response, replay is the wrong call.

## Hardware health

The remaining piece of #173's scope line. A Pi can hold its MQTT connection and heartbeat normally while its Modbus bus is unreachable — connection status and heartbeat both looked healthy, so a bank that could not open a single door was indistinguishable from a working one.

The heartbeat now carries `modbus_connected`, read from the state the bus actor already tracks. The field is optional at every layer: omitted when no bus is wired, `sometimes`-validated on the backend, and written only when the key is present. A client that reports nothing about its hardware must not be recorded as reporting the bus is down — those are different states, and `null` means the third one, never reported.

- `state-heartbeat.json` gains it as an optional boolean; existing clients stay valid
- `locker_banks` gains `modbus_connected` and `modbus_status_reported_at`
- `LockerBankResource` shows a Hardware badge (reachable / unreachable / unknown) beside the connection status, tooltipped with when it was last reported

## Still open on #173

- A response with `action: 'unknown'` is not routed by `CommandResponseReactor`, which handles `apply_config` and `open_compartment` only, so it will not clear a pending request. Needs the backend to have sent an actionless command, so it is narrow.
- Hardware logs ship to SigNoz but carry no `trace_id`: the poll path is deliberately untraced, and `inSpan` ends its span before the catch outside it runs. Searchable and alertable, just not linked to a trace.

## Verification

**locker-client:** `pnpm check` clean, 210/210 tests pass. 15 new tests cover the answered rejections, replay (including fresh `message_id` and the nothing-stored case), hardware logging, dropped-publish reporting, error-code mapping, and the heartbeat hardware field.

**locker-backend: not executed locally.** Docker was unavailable and the local `vendor/` predates the OpenTelemetry dependency, so `artisan test` fails before reaching a test. `php -l` and Pint pass on all changed files, and two handler tests are included, but CI is the first run that will actually execute them. Worth a reviewer's attention rather than assuming the backend half is proven.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `fix/190-admin-bugs` only — 27 files, +910/−70.

### Verified good

The contract change is additive and optional in both directions, which is the part most likely to go wrong in a fleet where devices update at their own pace. The schema marks `modbus_connected` optional with an explicit note, backend validation is `sometimes|boolean`, and — the detail that matters — **absent is not treated as false**: `array_key_exists` guards the write, so a device that reports nothing about its bus leaves the previous value alone rather than being recorded as broken. Older clients keep heartbeating unchanged.

It also closes the operational gap the issue described: a bank can be online and heartbeating while its Modbus bus is unreachable, and until now nothing could express that.

On the protocol side, the silent-drop complaint is properly addressed — `UNKNOWN_ACTION` and `MISSING_MESSAGE_ID` are now explicit codes and invalid commands get answered instead of vanishing. The duplicate-command path is a nice touch: the replayed response gets a *fresh* `message_id`, because the backend discards inbound duplicates by that field and a byte-identical replay would be swallowed.

On the issue's "without exposing credentials" requirement: the new rejection paths log structured fields (`transactionId`) rather than raw payloads, so nothing here widens the exposure surface.

### 1. `message.includes('modbus')` is too broad a catch-all

`isModbusLibraryError()` matches on library error text, which the docblock rightly flags as fragile and confines to one function. Four of the five patterns are specific transport failures — `port not open`, `econnrefused`, `timed out`, `crc error`. The fifth, `modbus`, matches essentially anything.

The consequence is concrete rather than theoretical. `isReconnectableModbusError` short-circuits on `error instanceof ModbusTransportError`, so every typed error the client raises itself bypasses the string matching — the catch-all only ever sees plain `Error`s. There is exactly one plain `Error` in the codebase whose message contains "modbus", and it is reachable from the command path:

```
yaml-config.repository.ts:137   throw new Error('modbus.port is required')
apply-config.ts:30              this.deps.config.reload()
apply-config.ts:39              catch → rollback → throw error   (rethrown unchanged)
command-dispatcher.ts:341       mapErrorToMqttCode(error)
errors.ts:59                    includes('modbus') → MODBUS_ERROR
```

So an `apply_config` whose reload fails validation reports **`MODBUS_ERROR`** to the backend — a hardware fault — while every other validation failure in that same use case throws `LockerError(INVALID_CONFIG)`. An operator is pointed at the wiring for a configuration typo.

`mapErrorToMqttCode`'s own docblock states the rule this breaks: "Errors we raise carry their own code… anything else is reported as unknown rather than guessed at from wording." Dropping the catch-all and keeping the four specific phrases costs nothing, because anything raised deliberately already carries its own code.

### Verdict

**Approve with one fix.** Dropping `includes('modbus')` is a one-line change that undoes a misattribution this PR would otherwise ship — a configuration fault reported to the backend as a broken bus.

PR #193, opened 2026-07-31
fix(backend): make command response dedup crash-safe (#188)

### LINKED ISSUE #188: Make backend command response dedup crash-safe
## Context

The locker client may replay the same semantic command response with the same `transaction_id` and a fresh `message_id`. Normal backend deduplication by `(locker_uuid, transaction_id)` handles this case.

## Verified findings

- **High risk:** `CommandResponseHandler` records the inbox row before persisting/dispatching `CommandResponseReceived`. A crash or event-persistence failure between those operations leaves the transaction permanently marked as seen without a corresponding domain event, so every replay is suppressed.
- Conflicting success/error replays are first-writer-wins, without payload/result conflict detection or logging.
- `CommandResponseDedupTest` repeats the same `message_id`, so it only exercises transport deduplication rather than a business replay with a fresh message ID.
- Parallel-insert and conflicting-result tests are missing.
- `command_transactions` has no retention or cleanup policy.

## High-level acceptance criteria

- Define and implement a crash-safe inbox/event boundary consistent with event sourcing.
- Test the same transaction with fresh `message_id` values, concurrent inserts, and conflicting results.
- Add observability for conflicting replays.
- Decide and document retention separately, or mark it as an explicit non-goal for this change.

Related: #171.

The implementation choice may affect API/MQTT reliability architecture and therefore may require a new ADR when that choice is made. This issue intentionally does not prescribe a solution.

Aibolit66:
Closes #188.

## The bug

`CommandResponseHandler` wrote the inbox row, then dispatched the domain event as a
separate step. A failure between the two left the transaction marked as seen with no
event behind it — every later replay was deduped away, and the response was lost
permanently with nothing to recover it from. Both writes now happen inside one
`DB::transaction()`, so the transaction is only marked as seen if the event landed too.

## Conflicting replays

Still first-writer-wins, but no longer in silence. A replay reporting a different
`result` or `error_code` now logs a warning carrying both outcomes — that case means the
device and the backend disagree about how the transaction ended, which is worth an
alert. Replays that differ only in bytes stay on the existing info line, with a
`payload_changed` flag for context.

## Payload hash

The hash now covers the response only. It previously included `message_id`, which is
fresh on every genuine replay, so any real replay looked like a payload that had
changed. The test output is what surfaced this.

## Tests

Three added, covering what the existing suite did not:

- the replay the transport guard cannot catch — same transaction, fresh `message_id`
  (the old test reused one `message_id`, so it only exercised transport dedup)
- a conflicting outcome — asserts the recorded result is untouched and the warning fires
- two callers racing the inbox — exactly one is told it recorded the transaction

Full backend suite: 306 passed.

## Deliberately out of scope

- **Retention for `command_transactions`** — still no cleanup policy. Left open rather
  than decided in passing.
- **Queue dispatch timing.** `after_commit` is `false`, so the queued job reaches Redis
  before the commit. If the commit itself fails, a worker can still process an event
  that rolled back. Closing that means flipping the queue setting globally or
  subclassing spatie's job class — both larger and riskier than this bug. The reactor's
  `derivedEventExists` checks are already idempotent, so the worst case is an orphan
  derived event rather than a lost response.
- **No ADR.** This is a transaction boundary around two operations that already existed.
  No API contract, MQTT topic, payload, or protocol change.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `fix/173-locker-client-error-observability` only — 3 files, +181/−10.

### Acceptance criteria — verified met

This issue had real criteria, and they are covered:

- **Crash-safe boundary** — inbox row and domain event now land in one `DB::transaction`, closing the "row survives a failed event write, every replay deduped away forever" hole.
- **Tests for all three requested cases** — `test_business_replay_with_fresh_message_id_is_deduplicated` (:309), `test_conflicting_replay_is_discarded_and_logged` (:347), `test_only_one_caller_is_told_it_recorded_the_transaction` (:403).
- **Observability for conflicting replays** — `Log::warning` carrying recorded *and* replayed result plus both error codes, so the disagreement is legible without a database dig.
- **Retention** — explicitly left open in the PR description, which is what the issue permitted.

The `payload_hash` change is the subtle one and it is right: excluding `message_id` and `traceparent` matters because a genuine business replay *always* carries a fresh `message_id`, so hashing them would make every legitimate replay look like a changed payload.

### 1. Queued reactors are dispatched inside the new transaction — *still live at tip*

`CommandResponseReactor` is `ShouldQueue` on the `events` queue, and all four connections in `config/queue.php` set `'after_commit' => false`. The event is now recorded *inside* `DB::transaction(...)`, so the reactor's job is enqueued while that transaction is still open.

Two ways that bites:

- A worker picks the job up before `COMMIT` and does not see the `command_transactions` row or the state the projector wrote.
- The transaction rolls back — the exact scenario this PR is about — and the job is already queued, so the reactor acts on a transaction that never happened, publishing MQTT for a response the backend has no record of.

That is the same failure class the PR fixes, relocated one layer out: before, an inbox row could outlive its event; now, a side effect can outlive its transaction. Setting `'after_commit' => true` on the connection, or deferring the dispatch to `DB::afterCommit()`, closes it.

### 2. `payload_changed` will be true for practically every benign replay — *still live at tip*

`Arr::except($payload, ['message_id', 'traceparent'])` deliberately drops the two delivery fields — but `timestamp` stays in the hash, and the adjacent comment says so itself: *"a re-sent response usually carries a fresh timestamp"*.

So the `payload_changed` flag on the info line is true almost every time a replay occurs, which makes it noise rather than signal. If it is meant to mean "the same outcome arrived with genuinely different content", `timestamp` belongs in the exclusion list alongside the other two. If it is meant to be informational only, it is worth saying so, because the name reads like an alarm.

### Verdict

**Approve with follow-ups.** Item 1 is worth resolving before this ships — it is a config-level change, but it reopens the durability gap the PR was written to close. Item 2 is cosmetic by comparison, though it costs one line.

One process note: retention was correctly declared out of scope, but it is declared only in the PR description. Once this merges that decision is not tracked anywhere — an issue would keep it from being silently forgotten.

Aibolit66:
## Independent review by AI

**Scope:** follow-up on the two findings. Static review; test figures are the author's.

**2 — fixed.** `timestamp` joins `message_id` and `traceparent` in the exclusion list, so `payload_changed` now means what its name says. Safe to loosen the hash because nothing load-bearing depends on it: `payload_hash` feeds only the info-line flag, while conflict detection compares `result`/`error_code` directly — a disagreeing replay is still caught and logged. The new test travels 30 seconds so the replay carries a genuinely fresh timestamp, then asserts the flag is `false`.

**1 — correctly deferred to #139.** Still live at tip: all four connections in `config/queue.php` are `'after_commit' => false` and `CommandResponseReactor` is still `ShouldQueue`. #139 item 2 already owns this, and flipping `after_commit` affects every queued job in the app — #139 also holds the `persist()` atomicity and the reactor idempotency race from #185, so fixing a third of the problem inside a dedup PR would be worse than leaving it whole.

Worth adding a line to #139 item 2 naming this PR as a concrete instance, the way item 3 names #185.

Retention remains declared out of scope in the PR description only, and untracked once this merges.

**Approve.**

PR #194, opened 2026-07-31
fix(admin): hide revoked and expired users from the group member list (#148)

### LINKED ISSUE #148: Remove Users from Group's User List if Expired/Removed
In the group edit UI, don't show those users at all instead of showing them as expired/removed. While it's nice to have the data, it's confusing to see it in that UI.

Aibolit66:
Closes #148.

## What was wrong

`MembersRelationManager` was bound to `Group::members()`, which is the raw pivot
relation with no filter. So the group edit screen listed everyone who had ever been a
member — including revoked and expired ones — and carried a "Removed at" column whose
only job was to explain why some of those rows did not count.

## What changed

`Group::activeMembers()` now mirrors `User::activeGroups()`, the relation the user-side
screen has been using all along: not revoked, and either no expiry or an expiry still in
the future. The relation manager points at it, so both directions of the same membership
finally agree on what "active" means.

Two things follow from that:

- The "Removed at" column is gone. Every row in the table is active by definition, so it
  would always be empty.
- `addableUserOptions()` no longer repeats the filter by hand — it reuses
  `activeMembers()`. A member who lapsed or was removed can be added back, which the
  picker already allowed but expressed as its own copy of the same query.

`members()` is untouched, so nothing that legitimately wants the full history loses it.

## Tests

- revoked and expired members are not visible in the relation manager table, active ones
  are
- a lapsed member is still offered by the add-member picker, so removing them from the
  table does not make them unreachable from this screen
- the existing smoke test now asserts the new relationship name

Full backend suite: 308 passed.

## No ADR

UI-level query filter reusing an existing relation pattern. No contract, schema, or
behaviour change outside this screen.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `fix/188-command-response-dedup-crash-safe` only — 3 files, +67/−13.

### Verified good

This does what the issue asked and takes the tidier route while doing it. Rather than filtering in the relation manager, it introduces `Group::activeMembers()` and points the relation manager at it — which also let the third copy of the same predicate disappear from `addableUserOptions()`. Net effect is less duplication than before, not more.

Removing the "Removed at" column is right and the comment says why: the table now only lists members who are neither revoked nor expired, so the column would always be empty.

Both genuinely risky paths are tested. `test_members_table_hides_revoked_and_expired_members` covers the visible behaviour, and `test_expired_and_revoked_members_can_be_added_again` covers the one worth going looking for — a previously revoked user is no longer in the table, so re-adding them must not collide with the existing pivot row. It goes through `GroupAccessService::addUser()` rather than a raw attach, and the test pins it.

Worth noting for the "we are hiding data" concern in the issue: the history is not lost. `UserAddedToGroup` and `UserRemovedFromGroup` are both presented in the audit log, so an admin asking "was this person ever in this group?" still has an answer — just not in this table, which is exactly what the issue wanted.

### 1. The "active membership" predicate now lives in two models — *still live at tip*

`Group::activeMembers()` and `User::activeGroups()` are byte-identical: same `wherePivotNull('revoked_at')`, same nested `whereNull('group_user.expires_at')->orWhere(... '>', now())`, including the same raw `group_user.` column references. The docblock even names the relationship — "the mirror of `User::activeGroups()`".

This PR reduced the copies from three to two, so it is moving the right way. But "active" is an authorization-adjacent definition: if it ever changes — a grace period after expiry, say — it has to change in both places, and a mismatch means the member list and the access checks disagree about the same membership, silently. A single shared scope or pivot-constraint helper would make that impossible rather than merely unlikely.

### Verdict

**Approve.** The one finding is a maintainability point, not a defect, and the PR left the code cleaner than it found it. No ADR needed — this is UI behaviour with no contract or architectural surface.

PR #196, opened 2026-08-03
feat(backend): bootstrap the first admin from configuration (#141)

### LINKED ISSUE #141: New First User/Admin Strategy
When we first deployed the project, we needed another way besides tinker/terminal to create our first admin. Back then we implemented the simple but unsecure version that enables registration whenever there are 0 users in the database and sets that user to being an admin (it might only set is_admin right now. Not sure if the new roles are being set yet?)

We should get rid of this system in favor of something more secure. My suggestion is having an environment variable for an admin email (like ADMIN_EMAIL) that can be set in coolify or wherever you're deploying. A user with the admin role and that email should be created on deployment if it doesn't exist/on first setup, and then the admin can simply reset the password for that email address when they first use the web app.

I'm open for other suggestions, but I think this one is pretty standard?

Aibolit66:
Closes #141. Includes `docs/adr/0034-first-admin-bootstrap.md`.

## The problem

Admin arrived by accident. Two mechanisms produced it:

- `AdminPanelProvider` mounted the Filament registration page while `User::count() === 0`
- `User::booted()` granted admin to the new user when `User::count() === 1`

Together: on a fresh deployment, the first person to find the instance became an
administrator.

The registration window was the *narrow* half — it closes by itself after the first
signup. **The hook was the real defect.** It granted admin as an invisible side effect of
`User::create()`, on any path: a seeder, a factory, an imported dataset, or a public
signup endpoint someone adds later. Nothing at the call site hinted at it, and the
condition it tested was a property of the whole table rather than of the user being
created. The database seeder already worked around it by calling `makeAdmin()` explicitly
— which is what the hook was supposed to make unnecessary.

## What replaces it

`ADMIN_EMAIL` names the administrator. `first-admin:create` creates the account and
grants the role through the existing event-sourced flow, so the grant is still auditable
with a null system actor.

It runs on deploy from an `entrypoint.d` script gated on `AUTORUN_ENABLED`. Every
container in the stack runs the entrypoint, but only the app container has that flag set
— the same flag that decides which container migrates — so the gate is what stops six
containers racing to create the same account.

The command is a **no-op** when an admin already exists, or when no address is configured.
That is deliberate: it runs unattended on every deploy, and failing over an
already-satisfied condition would take the container down. A malformed address is the one
real failure. It also refuses to be a general admin-maker — promoting a user belongs in
the panel, where it leaves an audit trail.

Removed: the registration toggle, the orphaned `Filament\Pages\Auth\Register` page, and
the `booted()` hook.

## The one risk worth knowing

The created account has a random password nobody ever sees, so the administrator gets in
through **password reset** — which means **mail has to work on a first deployment**. If it
does not, nobody can log in. That is exactly why `first-admin:create` also exists as a
manual escape hatch, and why `.env.example` says so next to the variable.

## Tests

Nine existing tests were quietly relying on the hook for their admin — the ADR predicted
this and it was the bulk of the work. They now grant admin explicitly, and the filler
users that existed only so the real subject would not be user number one are deleted.

One test asserted the removed behaviour directly (`first user becomes admin via an
event`). It is replaced by two: creating a user grants nothing, and a bootstrapped admin
is still granted through an auditable event with a null actor.

`CreateFirstAdminTest` covers the new command: configured address, explicit override,
no-op when an admin exists, idempotency, quiet success with nothing configured, failure
on a malformed address, and that the panel exposes no registration route.

`composer quality` green — Pint, PHPStan no errors, 317 tests.

## Note on the ADR

Written first and reviewed as part of this PR rather than separately, since #141 is one
issue. It records the alternatives — keeping the current window, artisan-only, a seeded
default password, a domain allowlist — and why each lost.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `docs/50-monorepo-release-strategy` only — 21 files, +1,508/−976.

### Verified good

The insecure path is genuinely gone, not merely discouraged. `AdminPanelProvider` no longer registers a registration page under any condition, the `User::count() === 0` branch and its `QueryException` workaround are deleted, and there is **no self-registration route in `routes/api.php` either** — so the old "first person to hit the site becomes admin" window is closed on both surfaces, not just the panel.

The bootstrap itself is careful in the places that matter:

- The generated password is `Str::password(32)`, never printed, never logged — the admin arrives through the reset flow.
- `adminRoleCount() > 0` restricts this to genuine bootstrap. The comment is right that a console command able to grant admin at any time would be a way around the panel's role flow and audit trail.
- Idempotent and safe on every container start, so it survives redeploys rather than needing a one-shot.
- The `AUTORUN_ENABLED` gate is the detail worth going looking for: every container in the stack runs the entrypoint, and without it six of them would race to create the same account on each deploy.
- `email_verified_at` is set outside the `firstOrCreate` because it is not fillable and would be dropped silently — with a stated rationale for why a verification click on top of the password reset would only add a second lockout path.

ADR-0034 covers the decision, which this change warrants.

### 1. A typo'd `ADMIN_EMAIL` takes the container down — *still live at tip*

The command treats an absent email as success, explicitly so — *"Failing here would take the container down over a missing optional setting."* But a **malformed** email returns `self::FAILURE`, and `50-first-admin.sh` runs under `set -e` with no guard on the call.

So the two misconfigurations behave in opposite ways: leave `ADMIN_EMAIL` unset and the app boots fine; set it to `admin@exmaple` and the container will not start at all — during a redeploy of a previously healthy system, over a setting the code otherwise treats as optional.

Failing fast on a typo is a defensible choice, but right now it reads as accidental rather than chosen, because the surrounding comment argues the opposite way. Either let the command warn and return success, or keep the hard failure and say in the script why a bad address is worth a failed deploy when a missing one is not.

### 2. Unrelated dependency work is bundled in — *marker not applicable*

`mobile-app/package.json` gains ~54 lines of `pnpm.overrides` — `minimatch`, `tar`, `undici`, `node-forge`, `brace-expansion`, `svgo`, `ajv`, `flatted` — with 1,858 lines of lockfile churn, plus changes to `locker-client/package.json` and its lock.

That is dependency-audit remediation. It has nothing to do with bootstrapping the first admin, and it is the majority of this diff by line count. It cannot be reviewed on its own merits here, and it cannot be reverted independently if one of those overrides turns out to break a build.

("Still live at tip" does not apply — this is about how the work was split, which a later branch cannot undo.)

### Verdict

**Approve.** The security substance is solid and it closes the hole the issue described. Item 1 is worth a decision before this reaches a deployment; item 2 is a note about commit hygiene, not something to unpick now.

PR #197, opened 2026-08-09
fix(backend): make last-admin role changes concurrency-safe (#187)

### LINKED ISSUE #187: fix(backend): make last-admin role changes concurrency-safe
## Context / problem

The last-admin guards currently check `hasOtherAdmin()` before recording a role revocation or deleting an admin. Two concurrent requests can both observe another admin and then remove both admin assignments, leaving the installation without an administrator.

This was identified while reviewing #157. The current UI exposes one selected role, while the event-sourced role model intentionally remains capable of multiple roles for future use.

## Scope

- Make admin demotion and deletion concurrency-safe.
- Preserve the event-sourced role audit trail.
- Cover simultaneous last-admin changes with a focused concurrency test.
- Keep normal role changes and future multi-role support unchanged.

## Proposed behavior

At least one active administrator must remain after any committed role-management or user-deletion transaction, including concurrent requests.

## Discussion / decisions needed

Choose the smallest reliable serialization boundary for the cross-user invariant, for example database locking or another transaction-scoped guard compatible with synchronous projectors.

## References

- PR #157
- Issue #142
- ADR-0021

Aibolit66:
Closes #187.

Stacked on #196 (`feat/141-first-admin-bootstrap`).

## What was wrong

The last-admin guards were check-then-act with nothing serializing them. `changeRole()`
called `User::hasOtherAdmin()` before recording the revocation, the model's `deleting`
hook called it again before a delete, and the bulk action compared `adminRoleCount()`
against the selection — but none of it held a lock or ran in a transaction.

On PostgreSQL's READ COMMITTED that is a straightforward TOCTOU. With admins A and B,
two simultaneous requests (delete A / delete B, or demote A / delete B) each read a
snapshot that still contains the other administrator, both pass their check, and both
commit. The installation is left with zero administrators and no way back in.

Single-request behaviour was never broken — demoting or deleting the last admin was
correctly refused. Only the interleaving was.

## What changed

`LastAdminGuard` wraps an admin-removing mutation in a transaction, takes a
transaction-scoped `pg_advisory_xact_lock` on one fixed key, runs the mutation, and then
asserts that at least one administrator remains — rolling back via `LastAdminException`
if not.

The postcondition is the important half. Checking afterwards, inside the serialized
transaction, is what makes the second request observe the first one's committed effect;
checking beforehand cannot. This is also exactly the invariant the issue asks for: at
least one administrator remains after any *committed* transaction.

A single advisory key rather than row locks, because the rows a request would lock are
precisely the ones a competing request may have just deleted, and one key cannot deadlock
against itself. SQLite serializes writers process-wide so the transaction is already the
boundary; other drivers fall back to ordered `lockForUpdate`.

Around that:

- `changeRole()` re-reads the target's roles *inside* the transaction. The old read
  happened before the check and could already be stale.
- `deleteUser()` / `deleteUsers()` on the service; the bulk one is all-or-nothing, so a
  selection can no longer be split into "some deleted, no admin left". It treats a falsey
  `delete()` as a violation, since the model's `deleting` hook vetoes by returning `false`
  rather than throwing and would otherwise commit as a silent no-op.
- The Filament `DeleteAction` and `DeleteBulkAction` route through the service via
  `->using()`. Their existing pre-checks stay as fast UI feedback.
- `DeleteBulkAction` now pins `->fetchSelectedRecords()`. It already defaulted to true,
  but both guards need real records — with it off, Filament deletes through a bare query
  and skips model events entirely.

## Tests

`LastAdminConcurrencyTest` — the invariant. Real threads cannot run against the in-memory
SQLite test database, so the competing request is replayed at the one point that matters:
while the lock is being acquired. Covers demotion, deletion, all-or-nothing bulk delete,
and that a refused change persists no stored event.

`LastAdminLockContentionTest` — the serialization itself, on a real second PostgreSQL
connection with a short `lock_timeout`. Skipped on other drivers, which is the 2 skips
below. A control test asserts the probe *can* take the lock when nothing holds it, so a
probe that always timed out could not masquerade as proof.

`UserDeleteActionTest` — the panel wiring, since `->using()` is the seam between the panel
and the service and would otherwise fail in the admin UI rather than in CI.

Each was checked against a deliberately broken implementation to confirm it fails for the
right reason: reverting the postcondition fails the demotion tests, stubbing out the lock
fails the contention test, and neutering the `->using()` closures fails the delete tests.

Full backend suite: 329 passed on SQLite (2 skipped) and 329 on PostgreSQL. Pint and
PHPStan clean. locker-client unaffected and green (210 passed).

## Dependency patches included

`docker-ghcr` was red on the audit steps for reasons unrelated to the fix above, so two
patch commits ride along (matching how the four audit fixes on #196 were handled):

- `league/commonmark` 2.8.2 -> 2.9.0. Six advisories published 2026-08-06 — five parsing
  denial-of-service issues plus an AttributesExtension unsafe-link filter bypass. It is
  transitive via `laravel/framework` (^2.8.1) and `laravel-lang/publisher` (^2.4.1), and
  2.9.0 satisfies both, so only the lock moves.
- `nanoid` 3.3.16 -> 3.3.18, reached through `vite > postcss`. The existing `postcss`
  override (`>=8.5.18`) was already satisfied by the locked 8.5.23 and so never moved;
  tightening it to `>=8.5.26` picks up postcss's own `nanoid ^3.3.17` constraint.

Both audits now exit clean, and the Docker build was verified locally against the updated
lockfiles.

## No ADR

ADR-0021 mentions the last-admin guards only as existing behaviour and never decided how
the invariant is enforced, so its decision is unchanged. The advisory lock is a reversible
implementation detail behind `LastAdminGuard`. If the pattern later gets adopted as the
house convention for cross-row invariants, that deserves its own ADR at the point there is
a second use case.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `feat/141-first-admin-bootstrap` only — 10 files, +615/−24.

### Verified good

The design is the right shape for the bug. `LastAdminGuard::protect()` runs the mutation inside a transaction and checks `adminRoleCount() < 1` as a **postcondition**, not a precondition — which is the actual fix, since the reported race is two requests each reading a snapshot that still contains the other's admin. A single fixed advisory key rather than per-row locks is also the correct call, and the docblock explains why: per-row locks cannot cover rows a concurrent request is in the middle of deleting, and one key cannot deadlock against itself.

The issue's other requirements are met. `test_a_losing_role_change_leaves_no_stored_events_behind` pins the event-sourced audit trail — a rolled-back attempt must not leave orphan events. `test_bulk_deletion_is_all_or_nothing` covers the multi-user path. `attempt()` alongside `protect()` keeps Filament's "show a friendly error" flow from having to catch exceptions.

### 1. The advisory lock is never exercised by any automated run — *still live at tip*

`LastAdminLockContentionTest` skips unless the driver is `pgsql`:

```php
if (DB::connection()->getDriverName() !== 'pgsql') {
    $this->markTestSkipped('The advisory lock only applies to the PostgreSQL driver.');
}
```

But `phpunit.xml` pins tests to `sqlite`/`:memory:`, and no workflow in `.github/workflows/` provisions a postgres service. So those two tests never run — not locally, not in CI.

Concrete evidence rather than inference: running `composer test` at the stack tip reports **`Tests: 371, Assertions: 1366, Skipped: 2`**, and `LastAdminLockContentionTest` contains exactly 2 tests.

That matters more than a usual coverage gap, because the `sqlite` branch of `acquireLock()` is `null` — it does nothing, relying on SQLite serializing writers process-wide. So `LastAdminConcurrencyTest` validates the *postcondition* logic on a driver that takes no lock at all, while the mechanism production actually depends on has no passing test behind it. The guard may well be correct; it just is not demonstrated.

A postgres service on the backend test job would turn these two from decorative into load-bearing.

Related, though lower stakes: the `default` branch falls back to `lockForUpdate()` on existing admin rows, which the comment concedes is "enough while at least one admin row exists" — i.e. weaker than the pgsql path against exactly this race. Only relevant if the project ever runs on MySQL.

### Verdict

**Approve with a follow-up.** The logic is sound and the event-sourcing requirement is properly covered. The finding does not block merge — but "concurrency-safe" is a claim that deserves a test that actually runs, and right now the two tests written to prove it are the two the suite skips.

PR #207, opened 2026-08-12
chore(backend): raise PHPStan to level 8 and fix the errors it surfaced (#199)

### LINKED ISSUE #199: Code review
Code-quality pass while there is a gap between feature tickets.

- [ ] Add `pcov` to `docker/8.4/Dockerfile` so `composer test:coverage` runs
- [ ] PHPStan level 6 → 7 (12 fixes)
- [ ] PHPStan level 7 → 8 (38 fixes, null-safety)
- [ ] locker-client: enable `noUncheckedIndexedAccess`, `exactOptionalPropertyTypes`, `noImplicitOverride` (21 fixes)
- [ ] Measure the same TypeScript strictness ladder for `mobile-app`

Out of scope: PHPStan level 9/10, raising the coverage percentage as a goal (81.9% is
healthy), and a PHPStan baseline.

Aibolit66:
Closes #199 (backend portion).

Stacked on #198 — review that first; the base branch is `feat/136-admin-edit-content-note`.

## What changed

Raises PHPStan from level 6 to **level 8** in `phpstan.neon.dist` and fixes everything that surfaced.

### Real bugs
- **`VerifyMosqHttpAuth`** — a request with `?mosq_secret[]=x` made `query()` return an array; the old `(string)` cast turned it into the literal `"Array"` (plus a PHP warning) before the `hash_equals` compare. Non-string input is now discarded up front.
- **`MqttTraceContext::extract()`** — missing null guard on the extracted context.
- **`CompartmentOpenRequestProjector::timestampOrNow()`** — returned a mutable `Carbon` while the app pins dates to `CarbonImmutable`, so the two branches disagreed on type. Now returns `CarbonImmutable`.

### Deduplication
- Extracted the duplicated Filament "Open" action into `app/Filament/Support/OpenCompartmentAction.php`. `CompartmentResource` and `CompartmentsRelationManager` now both call `OpenCompartmentAction::make()` — about 90 lines of copy-paste removed.
- New `Controller::authenticatedUser()` helper replaces `$request->user()` calls PHPStan could not narrow, in `AuthController` and `CompartmentController`.

### Typing only
`string[]` role names in `HasPermissions`, `array<int|string, string>` on the user-option pickers, `Closure` instead of `callable` in `AccessPickerOptions`, `(string)` casts where `__()` feeds a string parameter, and `@var` narrowing on relation accesses in five relation managers.

### Tests
- `MosquittoAuthControllerTest` — covers the array-secret case.
- `CompartmentOpenNotificationTest` — covers the open-notification path.

## Verification

Manual end-to-end run against the local stack and the locker simulator: opened a compartment from the Filament admin, watched the command reach the simulated bank over MQTT, closed the door in the simulator console, and confirmed the state came back through to the mobile app.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `feat/136-admin-edit-content-note` only — 21 files, +272/−122.

### Verified good

The claim most worth checking on a "level-raise" PR is whether a type fix quietly changed behaviour. The three places it plausibly could were checked, and it did not.

**The `OpenCompartmentAction` extraction is genuinely behaviour-preserving.** The PR says the two tables carried byte-identical copies, so the originals were diffed from the parent branch rather than taken on trust. The only differences were a comment and `\App\Models\User` written inline in one copy versus the imported `User` in the other — the authorization check, the session guard, the log context and all three notifications were identical. Bonus: the inline FQCN was against the project's own convention, and it is gone now.

**The Mosquitto auth change is a real security fix, not a cast-shuffling exercise.** `?mosq_secret[]=x` made `query()` return an array, and the old `(string) $providedSecret` turned that into the literal `"Array"` plus a PHP warning. Discarding non-strings is the right fix. The obvious hazard it could introduce was also checked — `$providedSecret` is now nullable and `hash_equals()` is strict-typed — and null is caught by the `if (! $providedSecret)` 401 branch before it ever reaches `hash_equals`. No TypeError path.

And it is *tested*, which is what makes this more than a static-analysis chore: `test_mosq_auth_rejects_an_array_shaped_secret` and `test_an_array_shaped_secret_cannot_stand_in_for_the_real_one` pin exactly the bypass shape. That second test — `?mosq_secret[0]=test-secret&mosq_secret[1]=x` — is the one someone would actually try.

**`HasPermissions::roleNames()` is safe.** The new `(string) $role` cast would be an `Error` if `UserRole` cast `role` to a backed enum; it does not — only `granted_at` is cast — so the values are plain strings and the cast is a no-op.

`Controller::authenticatedUser()` is a sound way to state the middleware's guarantee once: it fails loudly with a `RuntimeException` if a route is ever mounted without `auth:sanctum`, rather than degrading to a null-user path.

### 1. The blanket `ignoreErrors` pattern is now doing more work — *still live at tip*

`phpstan.neon.dist` keeps its one suppression:

```
'#Method .+::.+\(\) should return .+<.+> but returns .+<.+, \$this\(.+\)>#'
```

That was written against level 6. At level 8 the analyser reports strictly more, and a regex this loose — any method, any class, any generic return mismatch of that shape — will now silently absorb a wider set of findings than it was scoped for. It is the one place where "we are at level 8" overstates what is actually being enforced.

Worth checking what it currently suppresses and narrowing it to the specific classes that need it, so the level number means what it says.

### Verdict

**Approve.** No behavioural regressions found in the places most likely to hide one, and the level raise came with a genuine security fix plus tests for it. The single finding is about the honesty of the level number, not about the code.

Aibolit66:
Checked the one finding from the review — the blanket `ignoreErrors` pattern — by removing it and running the analysis.

It suppresses **32 errors, all of them the same Eloquent generics friction** it was written for: relation methods where Larastan expects `HasMany<X, Model>` and gets `HasMany<X, $this(Model)>`. Nothing unrelated is hiding behind it at level 8.

So the concern was theoretical rather than live. The pattern is loose enough that it *could* absorb an unrelated finding later without anyone noticing — worth narrowing to the specific classes if someone touches it — but it is not costing coverage today.

No change made.

PR #210, opened 2026-08-13
perf(mobile): cut Android internal build time where the profile says it goes (#204)

### LINKED ISSUE #204: Cache Android dependencies in local EAS builds
## Context

The Android job dominates the mobile build workflow. A recent successful run spent about 23 minutes in the local EAS Android build, compared with about nine minutes for the iOS build.

## Scope

- Profile the Android build to identify Gradle and EAS cache misses
- Add a safe cache strategy for generated native builds and Gradle dependencies
- Pin the EAS CLI version instead of resolving `eas-cli@latest` for every build
- Compare several warm and cold runs before and after the change
- Keep build reproducibility and signing behavior unchanged

## Notes

Cache keys should include the lockfile and relevant Expo/EAS configuration so native dependency changes cannot reuse stale outputs.

Aibolit66:
Closes #204.

Stacked on #207 — base branch is `chore/199-code-quality`.

## Measured result

Four `workflow_dispatch` runs, two per arm. Android job totals:

| Arm | Run | Time |
| --- | --- | --- |
| Before, cold (`dev`, unchanged) | `31632650060` | 23m10s |
| Before, repeat (`dev`, unchanged) | `31632696107` | 21m43s |
| **Before mean** | | **22m26s** |
| After, cold | `31634018022` | **14m23s (−36%)** |
| After, warm | `31634059986` | **10m25s (−54%)** |

Run-to-run variance on the unchanged config is 87s (6.3%), so both after-numbers sit far outside the noise.

Structural checks on the warm run, so the saving is attributable rather than inferred from wall clock:

- `x86` appears **zero times** in the log — the emulator ABIs genuinely are not built.
- **No lint tasks executed.** The only three `lintVital` mentions are echoes of the gradle command itself.
- **130 tasks resolved `FROM-CACHE`**, with ~1313 MB restored by the cache step in 13s.
- Remaining top tasks are `arm64-v8a`/`armeabi-v7a` CMake configure steps at 2–5s; the 103s-per-ABI Reanimated builds are gone.

## Why these changes and not the ones the ticket named

Profiling the last successful run (`28397897877`) contradicted the ticket's premise. Everything outside `eas build --local` totals **16 seconds**. Of 23m30s: ~1.9m pre-Gradle, ~21.6m Gradle, split as —

| Cost | Share |
| --- | --- |
| Per-ABI native C++ builds | ~6–7 min for `x86` + `x86_64` alone |
| `lintVitalAnalyzeRelease` across modules | ~2.5–3 min |
| **Downloads / toolchain installs** | **3–4 min — the entire share a dependency cache can touch** |

There is no R8/minification step in the build at all, so shrinking config was never a lever.

So: **build only device ABIs** (`-PreactNativeArchitectures=armeabi-v7a,arm64-v8a`), **skip release lint** (`-x :app:lintVitalRelease`), **enable the Gradle build cache** (`--build-cache` plus `actions/cache` over `~/.gradle/caches` and `~/.gradle/wrapper`, keyed on `pnpm-lock.yaml` + `app.config.ts` + `eas.json`), and **pin the EAS CLI** (`21.8.0`, replacing `eas-cli@latest` at three call sites — a release tool that changes without a commit is a reproducibility problem regardless of speed).

**Deliberately not done:** caching the generated `android/` prebuild directory, which the ticket proposed. Expo regenerates it from config every build, and a restored copy is the likeliest way to get a build that succeeds while being subtly wrong. The Gradle build cache gives the same benefit with input-hash correctness.

## Consequences worth reviewing

- **The APK no longer installs on x86/x86_64 emulators.** `mobile-app/docs/internal-builds.md` now says so — arm64 images (the Apple Silicon default) still work, otherwise build locally with `pnpm android`.
- **Android lint no longer runs anywhere in CI.** `mobile-app-ci.yml` lints TypeScript and never invoked AGP's `lintVital`, so this removes that coverage rather than relocating it. It was only running at artifact-build time on `main`, where nobody read it. If we want it back, the right home is a `lintRelease` step in `mobile-app-ci.yml` per PR — happy to add it here or as a follow-up.
- Signing is untouched: credentials still come from EAS at build time via `EXPO_TOKEN`.

## Note on the red checks

All four measurement runs are marked failed at the run level because **the iOS job fails**, for reasons unrelated to this change: the distribution certificate EAS signs with no longer exists in the Apple account, and the provisioning profile expired. It fails identically on unchanged `dev`. All four **Android** jobs passed. Worth its own issue.

Decision record: `docs/adr/0035-android-internal-build-abi-and-cache-strategy.md`.


Aibolit66:
## Independent review by AI

**Scope:** delta vs `chore/199-code-quality` only — 4 files, +151/−7.

### Acceptance criteria — verified met, including the one that is usually skipped

Every item in the issue is covered, and the measurement requirement — "compare several warm and cold runs before and after" — is actually done, with run IDs rather than adjectives: 23m10s before, 14m23s cold after (−36%), 10m25s warm after (−54%). That is the part most performance PRs assert and do not demonstrate.

The profiling is real too. ADR-0035 breaks the 23 minutes down by where it goes — the two emulator-only ABIs at ~6–7 minutes, `lintVitalAnalyzeRelease` at ~2.5–3, Gradle daemon and dependency work at 3–4 — which is what makes the fix targeted rather than hopeful. It also explicitly rejects dropping `armeabi-v7a` on the grounds that it would exclude 32-bit test hardware, so the ABI cut is scoped to what is genuinely unused.

Both consequences worth raising as findings are already documented as accepted trade-offs in the ADR: the APK no longer installs on x86/x86_64 emulators (with the workaround stated), and AGP's `lintVital` no longer runs anywhere in CI (with an explicit note that `mobile-app-ci.yml` covers a different class of problem). Both are deliberate, both are written down. `EAS_CLI_VERSION` pinning is done as asked.

### 1. `restore-keys` reintroduces exactly what the cache key was meant to prevent — *still live at tip*

The issue's constraint: *"Cache keys should include the lockfile and relevant Expo/EAS configuration so native dependency changes cannot reuse stale outputs."* The `key` does that correctly. But:

```yaml
restore-keys: |
  gradle-${{ runner.os }}-
```

On a lockfile change the exact key misses, and this prefix restores the newest cache from *any* previous key — i.e. outputs built against a different lockfile, which is the literal case the requirement rules out.

In practice this is fine and probably desirable: Gradle keys task outputs by input hash and recomputes what actually changed, so it is a warm start rather than stale artifacts. But it is worth stating that reasoning explicitly, because the workflow as written does not meet the requirement as written, and the next person comparing the two will have to re-derive why that is acceptable. It also means the "cold" number reflects an empty cache only if those measurements were taken with no prior cache entry at all.

### 2. Broken iOS signing is recorded only in this PR's description — *marker not applicable*

The PR notes that all four measurement runs failed at run level because the iOS job fails: *"the distribution certificate EAS signs with no longer exists in the Apple account, and the provisioning profile expired. It fails identically on unchanged `dev`."*

That is the iOS release path being down, not a measurement artifact — and it is the correct call not to fix it here. But there is no issue tracking it: nothing matching certificate, provisioning or signing exists in the issue list. Once this merges, the only record that iOS releases cannot be signed is a paragraph in a merged PR body.

Not a defect in this change. Worth a ticket before it is forgotten.

### Verdict

**Approve.** Best-evidenced PR in the stack so far — profiled before, measured after, trade-offs written into an ADR rather than left implicit. Item 1 is a comment-worthy clarification; item 2 is a ticket someone should open.

Issue #203, opened 2026-08-11
Run Expo health checks outside the main mobile CI path


merona-merlin:
## Context

`expo-doctor` is non-blocking but currently runs for every mobile-app change. A recent successful CI run spent about 16 seconds on this step out of roughly one minute total.

## Scope

- Run Expo dependency and health checks only when relevant dependency or Expo configuration files change
- Consider a scheduled run to retain regular ecosystem compatibility feedback
- Keep type checking, linting, formatting, and unit tests on every mobile-app change
- Preserve clear visibility when an Expo health check reports a problem

## Notes

This is a CI feedback-time optimization; it should not weaken the blocking mobile quality checks.

Aibolit66:
PR #217

Aibolit66:
Tried the cheap fix first — making `expo-doctor` a devDependency so `npx` stops re-downloading it each run — and **measured it: it does not help.** PR #217, now closed.

| | `npx expo-doctor` | local binary |
|---|---|---|
| Install dependencies | 9s | 10s |
| Expo doctor | **11s** | **12s** |

Same workflow, same runner type. No saving, slightly worse once the extra install is counted.

What this rules out: the time is **not** download or resolution overhead. It is the check itself. For comparison, `npx expo install --check` in the same job takes 1–2s, because `expo` is already a dependency — that contrast is what suggested the theory, and it was the wrong conclusion to draw, since the two commands do very different amounts of work.

So the ~12s is irreducible while the step runs on every mobile change, and the options left are the ones this issue originally proposed: path-filtering the step to dependency/Expo-config changes, and/or a scheduled run to keep ecosystem feedback.

Worth weighing against the Notes here: path-filtering means a compatibility break surfaces later and further from the change that caused it. Trading ~12s of feedback time for that is a judgement call, not a free win — the step is already `continue-on-error: true`, so it never blocks anything today.

PR #217, opened 2026-08-19
perf(mobile): resolve expo-doctor locally instead of downloading it per run (#203)

Aibolit66:
Addresses #203.

## Problem

The Expo doctor step ran `npx expo-doctor`, but `expo-doctor` was not a dependency — so every CI run downloaded the package before it checked anything.

The neighbouring step makes the cause unambiguous. Both use `npx`, in the same job on the same runner:

| Step | Command | Time |
|---|---|---|
| Expo dependency check | `npx expo install --check` | 1s |
| Expo doctor | `npx expo-doctor` | 11s |

`expo` *is* a dependency, so npx resolves it locally. `expo-doctor` was not, so npx fetched it. The cost was the download, not the check.

## Change

- `expo-doctor` added to `devDependencies`
- `"doctor": "npx expo-doctor"` → `"doctor": "expo-doctor"`
- workflow step `npx expo-doctor` → `pnpm run doctor`

Locally the check runs in ~5s with 18/18 checks passing. The CI figure for this branch is what the run on this PR will show against the 11s baseline above.

## Divergence from the issue

#203 proposed running the health checks only when dependency or Expo config files change, plus a scheduled run. This does something different, deliberately.

Path-filtering trades coverage for speed: the check stops running on most mobile changes, and a compatibility break surfaces later and further from the change that caused it. The issue's own Notes ask that the optimization not weaken the blocking checks. Removing the download gets the time back without making the check any rarer, so the filtering and the scheduled run are not needed.

Worth updating #203 to match, or closing it in favour of this.

## Trade-off

Pinning the version is a deliberate side effect: the tool no longer changes underneath CI without a commit. That is the same reasoning applied to `eas-cli@latest` in the Android build work. The cost is that newer checks arrive on a deliberate bump rather than silently.


Aibolit66:
Closing — measured, and it does not help.

| | Before (`npx expo-doctor`) | After (local binary) |
|---|---|---|
| Install dependencies | 9s | 10s |
| Expo doctor | **11s** | **12s** |

No saving, and marginally worse once the extra install cost is counted.

The premise in the description was wrong. It inferred the ~10s gap against `npx expo install --check` (1s) was download time, but those are different commands doing different amounts of work. The doctor check genuinely costs ~12s on CI hardware; a ~5s local run was a faster machine, not an absent download.

Findings recorded on #203. Nothing here is worth keeping: the version-pinning side effect is real but small, and it costs about a second per run to have.

PR #218, opened 2026-08-20
fix(locker-client): make shutdown and the async lifecycle safe (#170)

### LINKED ISSUE #170: Make locker-client shutdown and async lifecycle safe
## Context

Queue operations, reconnect delays, timers, and MQTT dispatch promises do not yet share a complete shutdown and error boundary.

## Scope

Define safe shutdown ordering, reject new work while closing, clean up timers, await active operations, and handle terminal async errors explicitly.

Aibolit66:
Closes #170.

## Problem

`shutdown()` cleared the timers and disconnected:

```js
clearInterval(pollTimer);
openCompartment.stopAllMonitoring();
heartbeat.stop();
await bus.disconnect();
await transport.disconnect();
await tracing.shutdown();
```

Nothing tracked commands already running — `dispatchMqttMessage` discarded its promise — so a redeploy mid-open could fire the relay, drop the transport before the response was published, and leave the request stuck at `sent` with a door possibly open. Nothing refused commands arriving during shutdown, a second SIGTERM started a second sequence, and a throw inside `shutdown()` never reached `process.exit`.

## Change

Ordering is now explicit: refuse new work → drain in flight → clear timers → flush pending responses → announce offline → disconnect transport, bus, tracing.

Two decisions worth naming:

- **The closing state is read at arrival**, not after parsing. Checking later refused commands that reached the client *before* shutdown began — the very work the drain exists to let finish. A test caught this.
- **Door detection is deliberately not awaited.** Its window can outlast the whole shutdown, so the pulse completes and `acknowledged` goes out, leaving the backend's timeout to cover a detection that never lands.

Timers are cleared *after* the drain, so nothing kills the monitoring belonging to a command still running.

**Explicit offline publish** covers what the Last Will cannot: a will only fires on an *ungraceful* disconnect, so a clean `docker stop` says nothing at all. This is **groundwork, not a behaviour change today** — `LockerConnectionStateHandler::handleValidated()` validates and logs, then stops at `// TODO Implement domain effects`, and `connection_status` is written by `LockerBankProjector` and the `DetectOfflineLockers` sweep. The payload is contract-valid and the message is now sent; acting on it is the handler's to implement. See ADR-0049.

**Each teardown step can be abandoned** rather than awaited forever. A wedged serial port would otherwise skip every later step and hand the process to SIGKILL in a worse state than giving up.

`stop_grace_period: 60s` in compose, since Docker's 10s default would kill mid-drain regardless of what the code intends.

## Error policy, stated explicitly

The ticket asks for terminal async errors to be handled *explicitly*, so the policy rather than just the code:

- **Transient** — a publish that failed, a poll that errored: log and continue, retried on the next tick. Already the behaviour in `publish()`, `pollOnce()` and `tick()`.
- **Terminal** — an unhandled rejection or uncaught exception: log and exit non-zero, let Docker restart clean. A process driving physical locks should not keep running in an unknown state.
- **A shutdown step that hangs or throws** — abandoned, sequence continues.

The simulator's dispatch boundary was the one genuinely unguarded async entry point and now matches `createApp`.

## No contract change

`error_code` is a free-form string in `response.json`, and `state-connection-lost.json` already accepts any `reason` — so `SHUTTING_DOWN` and `reason: "shutdown"` are valid today. The backend already consumes the topic via `LockerConnectionStateHandler`.

## Verification

270 tests, 269 pass, 1 skipped; lint, format and typecheck clean. 11 new tests cover the drain, the refusal, timer cleanup, a `disconnect()` that never resolves, a step that throws, a redelivery during shutdown replaying its stored answer, a refused command leaving no dedup record, and two concurrent deliveries of one transaction opening the door once.

Three review passes found three defects in the shutdown gate, each introduced by the fix for the previous one — all placement questions about where the refusal sits relative to duplicate detection and the in-progress claim. They are fixed in `c7ac1c44`, `8aafc325` and `82fa943e`; the invariant that keeps them from recurring (check and claim are atomic, refusal sits between the lookup and the claim) is now stated in the code and covered by tests.

Exercised against real Mosquitto and the backend using the fleet simulator:

```
← command  open_compartment #1 tx=refuse-f
Received SIGTERM, shutting down simulator...
→ response  open_compartment success tx=refuse-f      # in-flight command still answers
→ state/connection
← command  open_compartment #2 tx=refuse-d            # arrives during shutdown
Refused inbound MQTT command: shutting down
→ response  open_compartment error SHUTTING_DOWN tx=refuse-d
```

The backend did move to `connection_status=offline` after the clean stop — but that was the heartbeat sweep, not the new publish, which the handler does not yet act on. Correcting an earlier version of this description that attributed it to the publish.

**Not verified:** wedged serial port and Modbus timing on real hardware, and behaviour under an actual container `stop_grace_period` — all need a Pi.


Aibolit66:
Independent review found three things. Two were real defects, both now fixed in c7ac1c44; the third was a wrong claim in this description, corrected below.

**1. The refusal ran before duplicate detection.** A redelivered command that had already completed came back `SHUTTING_DOWN` instead of replaying its stored response. With a persistent session and QoS 1 that is a live case, and it is worse than the silence the gate replaced — the backend gets a confident failure for a transaction whose relay already fired. Validation now runs first, so a redelivery is answered from the store and only genuinely new commands are refused. Regression test added.

**2. The simulator leaked an unhandled rejection.** In-flight tracking used `void dispatched.finally(...)`, which builds a second chain off the raw promise while the caller's `.catch` guards a different one. A dispatch that rejects — a refusal whose own publish failed — became an unhandled rejection, and with no handler in `main-simulator.ts` that exits the process mid-shutdown, on the path this PR added. Now tracked via `.catch(...)`.

**3. Correction to this description.** The claim that the offline publish moves the backend to `connection_status=offline` is wrong. `LockerConnectionStateHandler::handleValidated()` validates and logs, then stops at `// TODO Implement domain effects`; `connection_status` is written by `LockerBankProjector` and the `DetectOfflineLockers` sweep. The `offline` state observed during verification was the heartbeat sweep — the mechanism the publish is meant to make unnecessary. The publish is correct groundwork and the payload is contract-valid, but it changes nothing today until the handler implements its domain effects.

Also adds **ADR-0049**, recording the shutdown ordering, refusal semantics, per-step abandonment and the transient/terminal error policy — the operability reasoning was previously only in this description, which disappears from view once merged.

Two follow-ups from the review, not addressed here:

- `await Promise.allSettled(inFlight)` is the only unbounded step in an otherwise bounded sequence, so a command that never settles still ends at SIGKILL. Recorded as a risk in ADR-0049 rather than fixed, since bounding the drain means choosing what to abandon.
- Backend-side: `CommandResponseReactor` checks only for an existing `CompartmentOpeningFailed` and not for a recorded `CompartmentOpenAcknowledged`, which is what made defect 1 damaging rather than merely wrong. Worth its own issue.

268 tests, 267 pass, 1 skipped; lint, format and typecheck clean.

Aibolit66:
Second review pass, both items addressed in 8aafc325.

**1. A refused command was recorded as work in progress.** Correct, and it came from the previous fix: moving validation ahead of the refusal also moved `markCommandInProgress` ahead of it, because `guardTransactionExecution` marked as its last step before returning `ready`.

The consequence is the one the review names — after restart, `recoverInterruptedCommands()` sweeps the `in_progress` record and publishes `UNKNOWN_ERROR`, giving the backend a second, contradictory answer for a command that already received a clean `SHUTTING_DOWN`, and a retry of that transaction id replays the stored error instead of executing.

`guardTransactionExecution` now only reports whether the command is a duplicate; marking moved to the call site, after the refusal check. The `requiresTransactionId()` guard mirrors the condition the dedup lookup was already made under, so handlers without a transaction id keep no records.

New test asserts both halves: the refused command leaves `getCommandRecord()` null, and `recoverInterruptedCommands()` then publishes nothing.

**2. Description corrected.** Lines 29 and 65 no longer claim the publish moved `connection_status`; they now say what ADR-0049 says — groundwork, with the observed `offline` attributed to the heartbeat sweep.

269 tests, 268 pass, 1 skipped; lint, format and typecheck clean.

Worth noting for whoever merges: the first fix traded one defect for a smaller one, and only a second independent pass caught it. Both were introduced by the shutdown gate rather than being pre-existing.

Aibolit66:
## Independent review by AI

**Scope:** delta vs `docs/adr-number-collisions` only — 9 files, +594/−15, in `locker-client/` plus ADR-0049.

### Verified good

The two non-obvious ordering calls are right: reading `closing` at arrival rather than after parsing is what stops commands that arrived *before* shutdown from being refused, and clearing timers after the drain avoids killing monitoring belonging to a command still running. `closeOrAbandon` does not leak when a step rejects after losing the race — `Promise.race` attached handlers on the way in.

The **"no contract change"** claim checks out: `error_code` and `reason` are both free-form strings in the schemas and in `LockerConnectionStateHandler::rules()`, so `SHUTTING_DOWN` needs no backend change.

The dispatcher gate went through three revisions, and the end state is the only one where all three properties hold at once — probed, not read:

```
refused command      -> relay 0, published SHUTTING_DOWN, dedup record: null
redelivered complete -> relay 1, stored success replayed (not contradicted)
concurrent same tx   -> relay 1
```

The last one is the one worth naming: an intermediate revision moved `markCommandInProgress` to the call site, which put two `await`s between the lookup and the claim and let two deliveries of one transaction both fire the relay. Keeping the claim inside `guardTransactionExecution` behind a `claim` flag restores the atomic check-and-claim while still recording nothing for a command that is about to be refused. The docblock says exactly that, which is the right place for it.

The simulator's `.catch(() => undefined).finally(...)` closes the unhandled-rejection leak while the caller still sees the original rejection.

ADR-0049 lands at a free number and records the ordering, the refusal semantics, the abandonment of stuck teardown steps, and the transient/terminal error policy — including the unbounded drain as an accepted consequence rather than an unstated one.

### 1. The PR body contradicts its own ADR — *marker not applicable*

ADR-0049 states the explicit offline publish is groundwork: `LockerConnectionStateHandler` only validates and logs, and `connection_status` stays driven by the heartbeat sweep. The PR description still says a clean stop "left the backend believing the bank was online until the heartbeat aged out" (line 29) and that "the backend moved to `connection_status=offline` after the clean stop" (line 65). The ADR is right; the description is what reviewers read first.

### Verdict

**Approve.** The shutdown gate now refuses new work, answers redeliveries from the store, and claims transactions atomically — with a test behind each. The only item left is two sentences in the PR body.

Aibolit66:
Third review pass found a regression from the previous fix, addressed in 82fa943e.

Moving `markCommandInProgress` to the call site put two awaits between the dedup lookup and the claim. Two deliveries of one `transaction_id` with different `message_id`s could then both read `ready` and both execute — the same door opening twice on one request, which is worse than the record-keeping problem that move was meant to solve.

The decision now travels into `guardTransactionExecution` as `claim`, so the lookup and the mark stay in one synchronous block and no await can open a window between them. `claim` is false only on the refusal path, so a refused command still records nothing for a relay that never fired, and a redelivery still replays its stored answer. Both earlier fixes hold.

The property this code has to keep — check and claim are atomic, and the refusal sits between the lookup and the claim — is now stated in the docblock and covered by a test that drives two concurrent deliveries of one transaction and asserts a single relay fire.

Worth recording for whoever merges: the shutdown gate produced three defects across three review passes, each one introduced by the fix for the previous. All three were placement questions about where the refusal sits relative to duplicate detection and the in-progress claim, and none were caught by the author's own testing.

270 tests, 269 pass, 1 skipped; lint, format and typecheck clean.

PR #219, opened 2026-08-20
fix(locker-client): complete runtime configuration handling (#174)

### LINKED ISSUE #174: Complete locker-client runtime configuration handling
## Context

Runtime config reload does not refresh every derived value, and validation and environment override behavior differ from the existing deployment contract.

## Scope

Refresh dynamic slave IDs, validate serial/MQTT/timer values, clarify environment overrides, and cover config reload and rollback behavior.

Aibolit66:
Closes #174.

Most of what the ticket describes was already handled by work since it was written on 14 July. Rather than assume that, each clause was checked — and where possible exercised against a live broker — before deciding what was left.

## Already resolved, verified rather than assumed

| Ticket clause | State | How it was checked |
|---|---|---|
| Reload refreshes derived values | works | Live: `apply_config` adding a compartment on a **new board** (slaveId 2) was accepted, and the retained snapshot came back `#1 #2 #3 #4` — polling followed the new mapping |
| Validate `apply_config` input | works | Live: duplicate relay target → `response apply_config error INVALID_CONFIG`. Schema also bounds `heartbeat_interval_seconds`, compartment numbers, slave ids and addresses |
| Config file presence and `modbus.port` | works | Throws at load |
| Flash duration bounds | works | Finite, min and max, enforced at load — the one setting that can cook a relay coil, so it was guarded first |
| Rollback on failed reload | works | Existing passing test: `apply config restores previous overlay when runtime reload fails` |
| Environment overrides | nothing to change | Env covers deployment (provisioning token, broker URL, credentials, paths, log level); the config file covers hardware and timing. The two sets are disjoint, and the code matches the documented deployment contract, so there is no precedence to resolve |

## What was actually missing

**1. Slave ids were captured, not read.** `WaveshareModbusBusActor` took its board list as a constructor argument and `reloadRuntimeConfig()` never refreshed it, so a board added by `apply_config` stayed invisible to the bus.

Latent rather than live: the only consumer is `runStartupFailsafe`, which runs at boot before any reload can arrive, and polling reads the config directly. But it is a trap for whoever adds the next caller. The list is now read when asked for.

**2. The validation asymmetry.** Commands arriving over MQTT are strictly checked. The operator-managed config file had almost none of that — only the flash duration, which is bounded because it can cook a relay coil. So `heartbeatInterval: 0` reached `setInterval` and published as fast as the loop allowed; `keepaliveSeconds`, `reconnectPeriodMs`, `connectTimeoutMs`, `maxReconnectAttempts` and every serial setting passed through unchecked, including non-numeric values.

The remote path being the strict one while the hand-edited one is not is backwards: the file is edited on a device with no feedback loop. Those values are now bounded at load, so a bad file fails at startup instead of becoming odd behaviour hours later. Sentinel zeros that mean "disabled" stay allowed.

**Serial values** are part of the same clause and are now covered: `baudRate` (1200–921600), `timeout` (50–60000 ms), and `dataBits` / `stopBits` / `parity` checked against the sets the driver's own types accept. `baudRate` matters most because it is not only handed to the driver — it feeds `calculateModbusRtuInterFrameDelayMs`, so a bogus value skews RTU bus pacing silently instead of failing, which is the opposite of what ADR-0035 exists to guarantee.

**The heartbeat bound says only what the contract says.** It is validated after the merge rather than on the base config, because the yaml key is deliberately ignored and the value only arrives through the runtime overlay — `apply_config` validates it inbound, but the overlay is a file on disk that can be edited or corrupted afterwards.

An earlier revision of this branch capped it at 300s. That was a number chosen for plausibility, and nothing upstream agreed with it: the inbound schema is `z.number().int().positive()` and the Filament field sets only `minValue(1)`. An admin could legally save 600, have it accepted and written, and then have the reload it triggered fail — recreating the asymmetry this PR set out to remove, inverted. A crash between the overlay write and the reload would also have left a file that threw on every boot, a crash loop on a device with no MQTT connection to correct it from.

The ceiling is now gone rather than widened: positive, whole, nothing more. The comment in the code explains why there is no maximum, since adding one back is the tempting move.

Nothing here touches flash duration.

## Verification

281 tests, 280 pass, 1 skipped; lint, format and typecheck clean. Ten new tests cover live slave ids, out-of-range and non-numeric transport and serial settings, framing values outside the driver's sets, sentinel zeros, a corrupted overlay heartbeat, a fractional heartbeat, and a heartbeat above the removed ceiling that must now load.

`locker-client/docs/modbus-configuration.md` gains the accepted range for every field — they were documented nowhere — and states that an out-of-range value fails at startup rather than being clamped, which matters for an existing deployment upgrading into stricter validation.

The reload and validation behaviour above was exercised against real Mosquitto and the backend using the fleet simulator, not only in unit tests.

### DIFF
--- a/locker-client/docs/modbus-configuration.md
+++ b/locker-client/docs/modbus-configuration.md
@@ -43,12 +43,20 @@ the required inter-frame silence from the configured serial parameters; see
 
 ## Base config fields
 
-| Field             | Required | Default | Description                            |
-| ----------------- | -------- | ------- | -------------------------------------- |
-| `port`            | Yes      | -       | Serial port path (e.g. `/dev/ttyACM0`) |
-| `baudRate`        | No       | 9600    | Serial baud rate                       |
-| `dataBits`        | No       | 8       | Data bits (7 or 8)                     |
-| `stopBits`        | No       | 1       | Stop bits (1 or 2)                     |
-| `parity`          | No       | `none`  | Parity (`none`, `even`, or `odd`)      |
-| `timeout`         | No       | 1000    | Response timeout in milliseconds       |
-| `flashDurationMs` | No       | 200     | Relay flash duration in milliseconds   |
+| Field             | Required | Default | Accepted            | Description                            |
+| ----------------- | -------- | ------- | ------------------- | -------------------------------------- |
+| `port`            | Yes      | -       | non-empty           | Serial port path (e.g. `/dev/ttyACM0`) |
+| `baudRate`        | No       | 9600    | 1200 – 921600       | Serial baud rate                       |
+| `dataBits`        | No       | 8       | 7 or 8              | Data bits                              |
+| `stopBits`        | No       | 1       | 1 or 2              | Stop bits                              |
+| `parity`          | No       | `none`  | `none`/`even`/`odd` | Parity                                 |
+| `timeout`         | No       | 1000    | 50 – 60000 ms       | Response timeout in milliseconds       |
+| `flashDurationMs` | No       | 200     | 100 – 500 ms        | Relay flash duration in milliseconds   |
+
+A value outside its accepted range **fails at startup** rather than being
+clamped or ignored, so a mistake surfaces at the moment of deployment instead of
+becoming odd behaviour hours later. `baudRate` is bounded because it also drives
+the RTU inter-frame delay — a wrong value there would skew bus pacing silently.
+
+Absent fields fall back to the default; only a value that is present and
+unusable is rejected.
--- a/locker-client/src/adapters/config/yaml-config.repository.ts
+++ b/locker-client/src/adapters/config/yaml-config.repository.ts
@@ -53,6 +53,43 @@ function parseBaseConfig(raw: unknown): BaseLockerConfig {
   };
 }
 
+/**
+ * Bounds a numeric setting read from the operator-managed config file.
+ *
+ * The MQTT path is already strict — `apply_config` is schema-validated before
+ * it reaches the client. The file is not, and it is the one edited by hand on a
+ * device with no feedback loop, so a typo there deserves the louder failure of
+ * the two. `heartbeatInterval: 0` would otherwise reach `setInterval` and
+ * publish as fast as the loop allows.
+ *
+ * Absent stays absent: the caller's default applies. Only a value that is
+ * present and unusable throws.
+ */
+function requireBoundedSetting(
+  value: number | undefined,
+  name: string,
+  min: number,
+  max: number,
+  alsoAllow?: number,
+): void {
+  if (value === undefined) {
+    return;
+  }
+
+  if (typeof value !== 'number' || !Number.isFinite(value)) {
+    throw new Error(`${name} must be a number`);
+  }
+
+  if (alsoAllow !== undefined && value === alsoAllow) {
+    return;
+  }
+
+  if (value < min || value > max) {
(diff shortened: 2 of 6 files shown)


Aibolit66:
## Independent review by AI

**Scope:** delta vs `fix/170-shutdown-async-lifecycle` only — 6 files, +258/−13, all in `locker-client/`.

### The ticket was mostly stale, and that was checked rather than assumed

#174 was written on 14 July, and the PR claims most of its clauses were overtaken by later work. That framing is the riskiest part of the change — it is where a real gap hides — so each clause was re-derived:

- **Reload refreshes derived values.** Flash duration is read per open, the heartbeat is explicitly restarted by `ApplyConfigUseCase`, and polling reads the config directly. Sound.
- **Validate `apply_config` input.** The inbound schema plus `validateCompartments` reject duplicate compartment numbers, duplicate relay targets and out-of-range addresses.
- **Config presence and `modbus.port`.** Throws at load — `port: ""` gives `modbus.port is required`.
- **Flash duration bounds.** `100` and `500` load, `501` throws. Untouched by this PR, correctly.
- **Rollback on failed reload.** Covered by an existing passing test, `apply config restores previous overlay when runtime reload fails`.
- **Environment overrides.** Genuinely nothing to resolve: env carries `PROVISIONING_TOKEN`, `MQTT_BROKER_URL`, `MQTT_CLIENT_ID`, credentials, `CONFIG_DIR`/`DATA_DIR`, `LOG_LEVEL`; the file carries `modbus.*` and the MQTT transport timers. No key appears in both, so there is no precedence to define.

The two gaps the PR identified are the two that were actually left. The slave-id fix is real and correctly called latent — `runStartupFailsafe`, its only consumer, runs at boot before any reload can arrive — and the validation asymmetry it names is the right one: the hand-edited file on a device with no feedback loop had less checking than the remote path.

### Findings and their resolution

**1. The heartbeat bound was narrower than the contract** — *already fixed at tip*. A 300s ceiling rejected values the admin panel legitimately offers (`->minValue(1)`, no maximum; inbound schema `positive()`), so a legal `apply_config` would be accepted, written, and then fail on the reload it triggered. The ceiling is gone rather than widened, which is the right shape: the file now says only what the contract says. `600` and `86400` load; `0`, `-5`, `1.5` and `soon` do not. This also closed the narrow crash-loop path where a dropped process between overlay write and reload left an unbootable client.

**2. Serial values were unvalidated, though the ticket names them** — *already fixed at tip*. `baudRate: fast` and `timeout: 0` used to load and reach the driver. Bounds now match the driver's own types (`7|8`, `1|2`, `none|even|odd`), and `baudRate` is bounded because it feeds `calculateModbusRtuInterFrameDelayMs`, where a wrong value skews RTU pacing silently instead of failing.

**3. Previously-valid files could stop the client booting, undocumented** — *already fixed at tip*. `docs/modbus-configuration.md` now carries an Accepted column for every field and states that an out-of-range value fails at startup rather than being clamped, with the `baudRate`/inter-frame reasoning attached.

`pnpm check` is clean and `pnpm test` reports 281 / pass 280 / skipped 1.

### Verdict

**Approve.** Every clause of #174 is now either implemented or shown to have been satisfied already, and the two genuine gaps are fixed without over-reaching — flash duration untouched throughout, which is the correct instinct for the one setting that can cook a relay coil.

Two cosmetic nits, neither worth blocking: the out-of-range message for the unbounded settings reads `must be between 1 and 9007199254740991`, which is what an operator sees on a Pi at boot; and the `mqtt.*` timer bounds are enforced but, unlike the modbus ones, still undocumented.

Issue #161, opened 2026-07-13
Introduce per-provisioning MQTT credential identities


merona-merlin:
## Context

MQTT device usernames currently equal the locker-bank UUID and are recreated with the same value after re-provisioning. Resetting a locker bank can delete the `mqtt_users` row, and `auth_opt_cache false` causes subsequent ACL checks to reach Laravel. However, if the same username is later recreated while an old MQTT session remains connected, that session may regain ACL rights without authenticating with the new password.

PR #120 remains an administrative recovery reset and will not solve this credential-lifecycle problem. It will delete the existing MQTT user and document that active-session termination is not guaranteed.

## Proposed direction

Issue a unique random MQTT username/credential identity for every successful provisioning generation. Never recreate a revoked identity.

Example:

```text
legacy: <locker-bank-uuid>
new:    <locker-bank-uuid>:<random-credential-id>
```

The exact username format is a decision, not a contract requirement. A fully random opaque username may be preferable.

## Why this removes the broker-kick requirement

After reset, the old `mqtt_users` identity is deleted or permanently disabled. Because a new provisioning receives a different username, the old active session's username never becomes authorized again. With go-auth caching disabled, future publish/subscribe ACL checks remain denied even if the TCP session itself stays connected.

A broker-side kick may still be useful for resource cleanup and immediate disconnect semantics, but it is not required for authorization correctness. This avoids maintaining a custom Mosquitto plugin solely for client eviction.

The assumption about ACL checks on already-connected clients must be verified against the pinned Mosquitto/go-auth versions in a real broker integration test.

### Backend

- Generate a unique MQTT credential identity for each provisioning.
- Store the identity in `mqtt_users.username` and map it to `mqtt_users.locker_bank_id`.
- Authorize device topics using the mapped locker-bank UUID instead of assuming `%u` equals the locker UUID.
- Return MQTT username and locker-bank UUID as separate provisioning response fields.
- Permanently revoke/delete the previous identity during reset.
- Ensure a revoked username is never reused.
- Pin and document the Mosquitto/go-auth versions whose ACL behavior is tested.

### Locker client

- Store MQTT username and locker-bank UUID separately.
- Use the MQTT username only for broker authentication.
- Use the locker-bank UUID for command, state, event, heartbeat, and response topics.
- Migrate the persisted credentials/config shape without breaking existing installations.

### Migration

- Existing UUID-based MQTT users remain valid as legacy generation-zero credentials.
[...]

PR #227, opened 2026-08-22
feat(mqtt): issue a new credential identity per provisioning (#161)

Aibolit66:
Closes the credential-lifecycle hole in #161. Decisions and alternatives: [ADR-0050](docs/adr/0050-per-provisioning-mqtt-credential-identities.md).

## The problem

A bank's MQTT username was its own uuid, and provisioning recreated it. Deleting the row revoked access, but recreating the *same username* for the next generation let a still-connected session pass ACL checks again without ever re-authenticating. Fixing that by evicting the session would have meant maintaining a Mosquitto plugin for client eviction.

## What changes

- **Opaque usernames.** Each provisioning mints 32 random characters. A retired username is never issued again, so there is no moment at which an old session becomes authorised — no broker kick needed.
- **Authorisation maps, it does not parse.** The ACL device branch resolves `mqtt_users.locker_bank_id` and matches topics against that uuid via a new `%l` placeholder, compared literally like `%u` so it cannot be injected as a pattern. `%u` is gone from the device patterns.
- **Retire, never delete.** Revocation sets `enabled = false` and stamps `revoked_at`; the unique index on `username` is what keeps the identity unusable. `deleteUser()` is removed.
- **Revoke before every insert**, inside the locked transaction — not only at reset. This reactor is queued and rethrows to retry, so without it a retry would leave a second live identity for the same generation.
- **`locker_uuid` in the provisioning reply.** The client stores it beside the username and builds every topic from it — command, state, event, heartbeat, response and Last Will — authenticating with the username alone.

## Migration

Legacy identities need no special case: their row already maps to its bank, so the same lookup authorises them until that bank is next reset. Credential files from an older client read `lockerUuid` as defaulting to `username`, which is what it was. Existing deployments need no action.

## Deployment ordering

**No bank may be re-provisioned until the updated client image is on the devices** — a legacy client would use an opaque username as its topic namespace. Backend and Pi images roll out separately, so this is a release-note gate, not something the code enforces.

**Restart the queue workers when deploying.** Issuance runs in a queued reactor; workers already running keep executing the old code and quietly keep minting uuid usernames. Provisioning still succeeds, which is what makes it easy to miss — it happened once while testing this.

## Verification

- Backend 393 tests, Pint, PHPStan clean; client 295 tests, format and lint clean.
- Broker image pinned to `2.1.0-mosquitto_2.0.15` — the build already running, so pinning changes only what a later pull may hand back.
- The premise is asserted, not assumed: with `auth_opt_cache false`, a connected session's publishes are refused on the next operation after its identity is disabled, with a new identity already issued. `locker-client/tests/integration/broker-revocation.test.ts` covers that sequence; it needs the stack, so it is opt-in behind `OPEN_LOCKER_BROKER_TEST=1` and skipped otherwise.
- Exercised end to end locally: a real reset and re-provisioning of a simulated bank issued an opaque identity, retired the previous one with `revoked_at`, stored username and locker uuid separately on the client, and the bank came back online.
- Reuse prevention is mutation-checked: restoring the old `firstOrNew` upsert makes `MqttUserServiceTest` fail.

## Not covered here

`mqtt_users` has no Filament surface, so retired identities accumulate invisibly and an operator cannot see which identity is live. Out of scope for #161, but worth a follow-up.

### DIFF
--- a/docs/adr/0050-per-provisioning-mqtt-credential-identities.md
+++ b/docs/adr/0050-per-provisioning-mqtt-credential-identities.md
@@ -1,0 +1,364 @@
+# ADR-0050: Per-provisioning MQTT credential identities
+
+## Status
+
+Accepted
+
+## Date
+
+2026-08-22
+
+## Context
+
+A locker bank's MQTT username is its own UUID. `MqttReactor` mints credentials
+with `$mqttUser = $event->lockerBankUuid`, and `MqttUserService::createUser()`
+upserts on that username with `firstOrNew(['username' => …])`. Resetting a bank
+deletes the row (`deleteUser((string) $lockerBank->id)`) and the next
+provisioning recreates the same username with a new password.
+
+That recreation is the problem. Deleting the row revokes access — with
+`auth_opt_cache false` every later ACL check reaches Laravel and is denied. But
+when the *same username* is created again for the next generation, an old
+session that is still connected starts passing ACL checks again, without ever
+having authenticated with the new password. The credential lifecycle has no
+boundary between generations.
+
+The obvious fix is to evict the old session at the broker. Mosquitto has no
+supported way to do that, so it would mean maintaining a custom plugin purely
+for client eviction — a lot of machinery for one edge.
+
+The authorisation path is also coupled to the username's meaning. The device
+branch of the ACL endpoint matches `locker/%u/state/#`, `locker/%u/response`,
+`locker/%u/event` and `locker/%u/command`, where `%u` is the authenticated
+username. The rule "you may touch your own locker's topics" is expressed as
+"your username *is* the locker UUID" — an assumption nothing enforces.
+
+Two things already in the codebase shape the answer:
+
+- **A generation identifier exists.** ADR-0048 added
+  `locker_banks.provisioning_generation`, and `MqttReactor::lockCurrentGeneration()`
+  already compares it with `hash_equals` to discard stale provisioning events.
+- **Disabling an identity is already honoured.** `mqtt_users.enabled` exists and
+  *both* the auth path and the ACL path already filter `->where('enabled', true)`,
+  and `username` already carries a unique index.
+
+## Decision
+
+**1. Every accepted provisioning issues a new, opaque MQTT username.**
+
+The username is 32 characters of `Str::random()`, carrying no structure. It is
+not derived from the locker UUID, the provisioning generation, or anything else
+a reader could parse.
+
+Opaque rather than `<locker-uuid>:<random>`: a parseable prefix invites exactly
+the coupling this ADR removes — the first `explode(':')` in an ACL path puts the
+assumption back, and nothing would fail loudly when it does. Operators correlate
+identity to bank through `mqtt_users.locker_bank_id`, which is what the mapping
+is for.
+
+**2. Authorisation maps identity → locker bank; it never reads the username.**
+
+The device branch of the ACL endpoint loads the enabled `MqttUser` (it already
+does), takes `$user->locker_bank_id`, and matches topics against that UUID as a
+literal. `%u` disappears from the device patterns.
+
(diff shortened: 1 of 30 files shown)


Aibolit66:
## Review trail

Five review passes ran against the ADR, the plan, and the implementation. Everything below was applied; nothing was carried forward unresolved.

### ADR (three passes)

| Finding | Applied |
|---|---|
| Rollout order contradicted decision 7 — opaque usernames were issued before the client shipped, the exact failure the Risks section names | Reordered: ACL → reply field → client → issuance, with the together-or-in-sequence constraint stated |
| "Never deleted" was not true as written: `locker_bank_id` is `cascadeOnDelete()`, and `deleteUser()`'s fate was unstated | Cascade accepted explicitly (a deleted bank has no successor generation); `deleteUser()` removal stated, then done |
| Decision 2 said "match the mapped uuid as a literal" without naming the mechanism, while leaning on `%u`'s injection safety | Specified the `%l` placeholder rather than interpolating the uuid into the pattern |
| `revoked_at` was written but never read, so the audit-trail argument outran the schema | Downgraded to a tombstone with a timestamp; attribution already lives in `LockerProvisioningReset` |
| Client seams cited from #161 no longer exist after the v2 rewrite | Named the real ones: `createApp.ts`, `createSimulatorApp.ts`, the credential store schema |
| No operability story for a revoked client that keeps running | Documented: refused auth, denials logged on the existing broker channel, plus a Negative consequence |
| Retained `state/connection` and Last Will from the old session unmentioned | Stated as harmless and overwritten on reconnect |
| **Insert-only + queued retry could mint two live identities for one generation** — `lockCurrentGeneration()` accepts the unchanged generation | Revoke-before-insert inside the locked transaction, making "one live identity per bank" an invariant; regression test added |
| Attribution paragraph read against the pre-revocation design; the converging-retry claim had an unstated tail | Both reworded; the tail (client holding credentials revoked by a retry whose replacement never arrived) is now named |

### Plan

| Finding | Applied |
|---|---|
| Pinning the broker *after* validating `:latest` proves nothing about what ships | Pin moved first; the assumption was verified against the pinned image |
| "Connect, revoke, assert denied" would pass under the old design too — it isn't #161's scenario | Test asserts the full sequence: connect → re-provision so a new identity exists → old session still denied |
| One commit does not satisfy a *deployment* ordering constraint | Release-note gate in the PR description |
| Contract change missed two files | `docs/asyncapi/examples/provisioning-success.json` and the client-side `provisioningSuccessResponseSchema` |
| `locker_uuid` should be optional client-side | Optional with a `mqtt_user` fallback, so a rolled-back backend still works |
| `revokeForLockerBank()` scoping | Scoped to `enabled = true`, stamps `revoked_at` in the same update, called inside the `lockForUpdate` transaction only |

### Implementation

| Finding | Applied |
|---|---|
| `…from contract-shaped success reply` was misleading — the schema makes `locker_uuid` required, and that test omits it deliberately | Renamed to `…from a pre-locker_uuid backend reply` |
| Integration test could strand fixtures: the bank was created before the `try`, so a failure in between skipped cleanup | Cleanup registered with `t.after()` before anything is created and keyed by name, covering partial setup failures |
| `LockerBank::create()` bypasses the aggregate, so a leaked fixture would appear with no event stream | Docblock says so, and says not to copy the pattern |
| Nothing pinned where `locker_uuid` comes from — a wrong source would still pass schema validation | `assertSame($event->lockerBankUuid, …)` plus `assertNotSame` against `mqtt_user` |
| **No test held `createUser()` to failing on a reused username** — the property the security argument rests on; restoring `firstOrNew` would have left the suite green | `MqttUserServiceTest` added, and mutation-checked: restoring the upsert makes it fail |

### Found by running it, not by review

The first end-to-end attempt silently ran the **old** code and reissued a uuid username: the queue workers had booted before the change. Provisioning succeeded, which is what made it easy to miss. Recorded in the ADR rollout and in this PR's deployment notes.

Issue #172, opened 2026-07-14
Recover locker-client Modbus after prolonged disconnects


merona-merlin:
## Context

The current reconnect attempt budget can remain exhausted until process restart, and initial serial-port failures are not consistently retried.

## Scope

Make reconnect cycles recoverable after long USB or board outages, classify serial errors reliably, and add disconnect/reconnect tests.

Aibolit66:
PR #228

Issue #169, opened 2026-07-14
Run Raspberry Pi locker-client cutover soak test


merona-merlin:
## Context

Mac hardware tests passed with three Modbus boards, but the final Raspberry Pi runtime, reconnect, MQTT, and long-running behavior still needs validation before production use.

## Scope

Run a multi-board soak and failure-recovery test on the target Pi, record results, and close remaining hardware validation from #97 and ADR-0029.

PR #228, opened 2026-08-22
fix(locker-client): recover the Modbus bus after a prolonged outage (#172)

Aibolit66:
Closes the recovery hole in #172. Decisions and alternatives: [ADR-0051](docs/adr/0051-modbus-reconnect-declares-an-unreachable-bus.md), which **partially supersedes [ADR-0006](docs/adr/0006-best-effort-startup-failsafe-for-unreachable-modbus-boards.md)** for the unreachable-bus case.

## The problem

`ReconnectCoordinator` allowed five attempts and reset `attempts` only on a successful connect. Twenty-five seconds with the adapter unplugged left the counter spent, and every later attempt was refused immediately — a Pi that lost its adapter for half a minute stayed unable to reach its hardware until the container restarted, however long the adapter had been back.

Three faults sat alongside it: startup got a single attempt because `connect()` bypassed the coordinator; reconnectable errors were recognised by message text, so an unplugged adapter (`ENOENT`/`EIO`/`ENXIO`) was classified as permanent; and `ConnectionState` had no way to say "dead", leaving a spent bus reporting `connecting` forever.

## What changes

- **The budget belongs to the outage, not the process.** A spent cycle is retried once a cooldown passes, and the periodic compartment poll starts the new cycle — no timer of its own. Elapsed time is measured **monotonically**: a Pi has no RTC, and a backward jump at first NTP sync would have re-created the latch this removes.
- **A spent cycle declares `unreachable`,** in **both** failure paths — the command path via `ensureConnected()` and the poll path via `runWithReconnectRetry()`, which is the one that notices a bus dying between commands at all.
- **All three dial sites go through the coordinator** — `connect()`, `ensureConnected()`, `reloadRuntimeConfig()` — so startup and config reloads get a cycle rather than one attempt.
- **Startup survives a missing adapter.** `runStartupFailsafe` skips its sweep when the bus is unreachable, while still failing when the bus is reachable and every board stays silent — a wiring fault a human must fix. Previously that throw reached `main().catch(process.exit(1))`, restarting the process until the adapter returned.
- **Classification by error code**, read from the error and a nested `cause`: `ENOENT`, `ENXIO`, `EIO`, `EBADF`, `ECONNREFUSED`. `EACCES` is deliberately excluded — a missing `dialout` group fails that way every time, and cycling against it is the loop this exists to avoid. `isModbusLibraryError` gained the same codes, so an unplugged adapter reports `MODBUS_ERROR` rather than `UNKNOWN_ERROR`.
- **New setting** `modbus.reconnectCooldownSeconds` — default 60, bounds 5–3600, validated like `mqtt.keepaliveSeconds`. Optional; existing deployments inherit the default.

## Nothing changes on the wire

`modbus_connected` is derived from `state === 'connected'` and was already `false` while a dial was in progress, so the admin's Hardware badge reads the same before and after. `unreachable` is for the client's own correctness — distinguishing a spent cycle from an attempt in flight, in the logs and in the code that decides whether to start another.

## Worth knowing when reviewing

- `markUnreachable()` only fires when `isCycleSpent()`, which requires `maxAttempts > 0`. The actor always passes 5, so nothing today hits it — but a bare `ReconnectCoordinator` (default 0, unlimited) would never report `unreachable`. Latent assumption, not a bug.
- `readDoorSensors` swallows failures into `'unknown'` by design, so the poll path cannot signal a dead bus by throwing. That is precisely why the state carries the verdict.
- `createSimulatorApp.ts` calls the failsafe too, but `InMemoryLockerBus` never reports `unreachable`, so the behaviour change does not reach the simulator.

## Verification

309 tests pass, 2 skipped, 0 fail; format, lint and typecheck clean; the simulator runs unaffected. 14 tests added, including the poll-path case, the return-to-`connected` case, once-per-cycle log volume, and the failsafe split.

**ADR status stays `Proposed`.** The error-code list is what the serial layer is documented to report for a removed device — it has not been confirmed against a real adapter being unplugged. That belongs to #169's soak test, and a code guessed wrong means no reconnect for that failure.

### DIFF
--- a/docs/adr/0006-best-effort-startup-failsafe-for-unreachable-modbus-boards.md
+++ b/docs/adr/0006-best-effort-startup-failsafe-for-unreachable-modbus-boards.md
@@ -79,6 +79,15 @@ could continue.
   reset until connectivity is restored
 - polling is less noisy during startup because it begins on a reachable board
 
+## Supersedes / Superseded By
+
+- Superseded in part by
+  [ADR-0051](0051-modbus-reconnect-declares-an-unreachable-bus.md): the rule below
+  that startup fails when every configured board fails the failsafe no longer
+  applies when the Modbus bus itself is unreachable, which is recoverable and is
+  handled by reconnect instead. It still applies when the bus is reachable and the
+  boards are silent. Everything else here stands.
+
 ## References
 
 - `docs/adr/0004-waveshare-hardware-flash-and-supported-boards.md`
--- a/docs/adr/0051-modbus-reconnect-declares-an-unreachable-bus.md
+++ b/docs/adr/0051-modbus-reconnect-declares-an-unreachable-bus.md
@@ -1,0 +1,288 @@
+# ADR-0051: Modbus reconnect declares an unreachable bus and keeps recovering
+
+## Status
+
+Proposed
+
+## Date
+
+2026-08-22
+
+## Context
+
+`ReconnectCoordinator` allows five attempts, and `attempts` is reset only by a
+successful connect. After five consecutive failures — roughly twenty-five seconds
+with a USB adapter unplugged — the counter stays at five, and every later attempt
+short-circuits on `attempts >= maxAttempts` and throws immediately. Nothing ever
+lowers it again.
+
+So a Pi that loses its adapter for half a minute stays unable to reach its
+hardware until the container is restarted, however long the adapter has been back.
+Restarting is the only recovery, and nobody is standing next to the locker.
+
+Three smaller faults sit alongside it:
+
+- **Startup gets one attempt.** `connect()` calls `connectInternal()` directly
+  rather than through the coordinator, so a port that is not ready at boot is
+  never retried, while the same failure a minute later is.
+- **Reconnectable errors are recognised by message text.**
+  `isReconnectableModbusError()` falls back to `message.includes('Port Not Open')`
+  or `'ECONNREFUSED'`. An unplugged adapter surfaces as `ENOENT`, `EIO` or
+  `ENXIO`, none of which match, so the most recoverable failure of all is
+  classified as permanent and no reconnect is attempted. #173 removed the
+  mirror-image fault — a substring that matched too much — from the same layer.
+- **The state machine cannot express "dead".** `ConnectionState` is
+  `disconnected | connecting | connected`, and `connectInternal()` sets
+  `connecting` before dialling. When the attempt fails the state stays
+  `connecting`, so a permanently unreachable bus is indistinguishable from one
+  that is mid-attempt.
+
+What already works is the reporting path: the heartbeat carries
+`modbus_connected` (`state === 'connected'`), the backend stores it, and the admin
+panel renders a "Hardware" badge of reachable / unreachable / unknown. An operator
+can already see a dead bus. It is the client's own notion of its state, and its
+ability to recover, that are wrong.
+
+ADR-0014 faced the same question for the MQTT link and chose unlimited automatic
+reconnects, with a finite cap available as "an optional safety cap for lab use".
+The two transports in the same process currently disagree about whether
+permanent surrender is acceptable on hardware in the field.
+
+## Decision
+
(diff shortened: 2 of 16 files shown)


Aibolit66:
## Review trail

Four review passes — two on the ADR, two on the plan, one on the implementation. Everything below was applied; nothing carried forward.

### ADR (two passes)

| Finding | Applied |
|---|---|
| Decision 2 overstated what reaches the operator — `modbus_connected` is one boolean, already `false` during a dial, so the badge reads the same before and after | Rewritten to say plainly that nothing changes on the wire and that `unreachable` is for the client's own correctness |
| `recoveryCooldownMs` was introduced with no default, home or bounds, while Rollout said "no configuration changes" and Risks said "it is configurable" | Named: `modbus.reconnectCooldownSeconds`, default 60, bounds 5–3600, validated like `mqtt.keepaliveSeconds`; both sections reconciled |
| `reloadRuntimeConfig()` has the same single-attempt fault decision 4 fixes, but only `connect()` was named | Both dial sites named in the decision and in Rollout |
| Which layer sets `unreachable` was unspecified, and the path that matters is the quiet one | Both failure paths named, with why the poll path matters more — wiring only `ensureConnected()` would undercut decision 1's own recovery trigger |
| Rollout drifted from the Decision edits (setting name, missing failsafe step, missing tests) | Rollout rewritten to match, twice |
| Alternatives didn't cover extending the heartbeat with a reason field | Added as Alternative E, with its own revisit trigger |
| Narrowing an accepted ADR needs the lifecycle bookkeeping, not just prose | ADR-0051 "partially supersedes" ADR-0006; ADR-0006 carries a back-pointer |
| Decision 5 still listed `EACCES` after the plan dropped it, and `ECONNREFUSED` sat under "existing string cases" after being moved to the code list | Both corrected, with `EACCES`'s absence justified rather than silent |
| `config.yml` is really `locker-config.yml`; the 5–3600 bounds come from `mqtt.keepaliveSeconds`, not the Modbus settings | Both fixed |

### Plan (two passes)

| Finding | Applied |
|---|---|
| **Making `connect()` non-throwing doesn't make boot survive a missing adapter** — `runStartupFailsafe` throws when no board answers, and that reaches `main().catch(process.exit(1))` | Failsafe skips its sweep when the bus is unreachable; the zero-success throw stays for the reachable-bus case, and ADR-0051 records the narrowing |
| `mapErrorToMqttCode` would report `UNKNOWN_ERROR` for the very fault being fixed, since `isModbusLibraryError` matches on message text | Extended with the same code list |
| `ECONNREFUSED` is a code, not a message — leaving it as a substring keeps the pattern decision 5 argues against; and a code on a nested `cause` is unreachable | Moved into the code list; `serialErrorCode()` walks `cause` |
| Missing tests: once-per-cycle log volume, exit from `unreachable`, config bounds | All three added; cooldown tests use a small `cooldownMs` rather than a fake clock, matching how the existing tests drive `delayMs` |

### Implementation

| Finding | Applied |
|---|---|
| **The cooldown used `Date.now()`** — a Pi has no RTC, and a backward jump at first NTP sync makes the elapsed check negative, latching the bus until wall-clock catches up. Most likely at boot, which is when a missing adapter is most likely | `monotonicNow()` wrapping `performance.now()`, with the reasoning in the code |
| "60 seconds" encoded in three places | One home: `DEFAULT_MODBUS_RECONNECT_COOLDOWN_MS`; the actor and `createApp` pass through rather than restating it |
| The `isModbusLibraryError` docblock — including the #173 history — was orphaned above a constant | Moved back above the function, with one clause updated since message text is no longer its only signal |

### Found by running the tests, not by review

Two of my own test assumptions were wrong, and both revealed why the design is what it is: `readDoorSensors` swallows failures into `'unknown'`, so the poll path cannot signal a dead bus by throwing — which is exactly why the state carries the verdict; and `ensureConnected()` short-circuits on `driver.isOpen()`, so a fake that "unplugs" without closing the handle never reaches the reconnect path at all.

PR #258, opened 2026-08-25
fix(mobile): refresh terms acceptance without restarting the app (#250)

### LINKED ISSUE #250: User ToS acceptance refresh
Whether a user has accepted terms does not refresh via reverb OR via swipe-down refresh. Restarting the app helps and logout/login works. This means that if an admin publishes new ToS, the user will get 403s without knowing what's happening and they won't see the banner or prompt that tells them to accept the new ToS until they restart the app or log out and back in.

Optimally, we want reverb to make sure the user gets a live update about the terms, and if reverb doesn’t work for some reason, swiping down to refresh should get the updated state of their ToS acceptance (and maybe some other user data that currently might turn stale).

What does work is the user no longer having access (shows 403) when the new terms are published, so at least security is fine.

Aibolit66:
Closes #250. Reasoning and alternatives: [ADR-0056](docs/adr/0056-account-state-on-its-own-realtime-channel.md).

## What was wrong

`terms_current_accepted` comes from the user profile, which the app caches and nothing refreshed. When an admin activated new terms the app kept believing the old answer: the prompt to accept never appeared, and every write came back 403 with nothing to explain it. Restarting the app or signing out again was the only cure.

## Four layers, cheapest first

The issue asks for a realtime update with pull-to-refresh as the fallback. Both are here, plus a third path that removes the symptom entirely.

1. **The refusal explains itself.** A 403 carrying `code: terms_not_accepted` invalidates the cached profile, so the very request that was refused produces the prompt explaining why. Deterministic, no socket involved. Matched on the code, never the message — the message is localised, and a 403 that merely mentions terms is not the gate refusing.
2. **The existing fallback now covers it.** `refetchFallback` in the realtime hook already ran on socket drop and on returning to the foreground; it invalidates `Auth` alongside `Compartment`. That covers the "restarting helps" complaint using machinery that was already there.
3. **Pull-to-refresh refetches the profile**, not just compartments — the gesture people reach for when something looks wrong.
4. **Reverb makes it live.** `TermsAcceptanceRequired` on a new per-user channel; the app invalidates `Auth` on receipt.

Layers 1–3 are what make a missed broadcast a delay rather than a defect.

## Contract added

| | |
|---|---|
| Channel | `users.{id}.account`, private, authorised like the compartment channel |
| Event | `.terms.acceptance-required` (leading dot: no Echo namespacing) |
| Payload | `{ version: number }` — a signal, not content |

Three decisions worth knowing, all argued in the ADR:

- **Activation, not publication.** Acceptance is compared against the *active* version, so that is the moment a cached profile becomes wrong. One service call records both today, but an activate-an-existing-version path would otherwise broadcast nothing.
- **Its own channel**, not the compartment one. Overloading `compartment-status` would have been two lines with no new plumbing — and would make the channel name stop describing its contents.
- **One Echo instance** for both channels. A hook per channel means a websocket per channel, for an event that fires perhaps twice a year.

## Verified

- Backend: 404 tests, PHPStan clean, Pint clean. Two new tests cover the dispatch and the channel/payload shape.
- Mobile: 26 tests (was 19), 18/18 checks. `termsGate.test.ts` covers the 403 predicate against the real body, a different 403 code, a message-only impostor and non-object bodies; `accountChannel.test.ts` pins the channel and event names against the backend's route and `broadcastAs()` — they are string literals on two sides of a network boundary and drift fails silently.
- **End to end against the running stack**: published a version in the admin panel and watched the prompt appear in the app without touching it.

One thing that cost time and is worth repeating: the first attempt showed nothing because the **event-worker had the old reactor in memory**. Queued reactors need a worker restart on deploy, same as the credential-issuance change in #227.

### DIFF
--- a/docs/adr/0056-account-state-on-its-own-realtime-channel.md
+++ b/docs/adr/0056-account-state-on-its-own-realtime-channel.md
@@ -1,0 +1,120 @@
+# ADR-0056: Account-level realtime state gets its own per-user channel
+
+## Status
+
+Accepted
+
+## Date
+
+2026-08-25
+
+## Context
+
+[ADR-0023](0023-mobile-realtime-compartment-status-via-reverb.md) put compartment
+state on a private per-user Reverb channel, `users.{id}.compartment-status`. It is
+the only channel the app subscribes to.
+
+Terms acceptance needed the same treatment. A user's profile carries
+`terms_current_accepted`, the app caches it, and nothing told the app when new
+terms were activated — so the prompt to accept never appeared and every write
+failed with a 403 the app could not explain (#250).
+
+The obvious shortcut was to broadcast that on the existing channel. It is already
+authorised in `routes/channels.php`, the app is already subscribed, and Echo is
+already connected. Adding one more `listen()` would have been a two-line change
+with no new plumbing at all.
+
+## Decision
+
+Account-level state broadcasts on its own private channel, `users.{id}.account`,
+authorised the same way as the compartment channel.
+
+A channel is named after what it carries. `compartment-status` carrying "your
+terms are out of date" makes the name a lie, and a name that has lied once stops
+being read — the next unrelated event goes on whichever channel is nearest, and
+the app ends up filtering a firehose to find what it cares about.
+
+Both channels are served by **one Echo instance**. A hook per channel would open a
+websocket per channel, and this event fires perhaps twice a year.
+
+The payload is a signal, not content: `TermsAcceptanceRequired` carries a version
+number and nothing else. The app invalidates its cached profile and re-reads it,
+so there is exactly one answer to "must I accept" and it comes from the API.
+Broadcasting the terms themselves would put a second copy on the wire, free to
+disagree with the first.
+
+Broadcasts hang off **activation**, not publication. `terms_current_accepted`
+compares a user's acceptance against the *active* version, so activation is the
+moment the cached profile becomes wrong. Today one service call records both, but
+an activate-an-existing-version path would otherwise broadcast nothing.
+
+## Rationale
+
+The realtime path is not the only defence, and should not be. #250 also added a
+403-triggered refresh and extended the existing reconnect/foreground fallback, so
+a user whose socket is down still recovers. This channel is the fast path, not the
+guarantee — which is why a missed event costs latency rather than correctness.
+
+## Alternatives Considered
+
+### Alternative A: Broadcast on `users.{id}.compartment-status`
+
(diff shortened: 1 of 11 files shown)


Aibolit66:
## Review trail

Two independent review passes over the plan and two over the implementation. Everything below was applied.

### Plan

| Finding | Applied |
|---|---|
| **`refetchOnFocus`/`refetchOnReconnect` would have been no-ops** — `setupListeners` binds browser focus/online events, which never fire in React Native | Dropped in favour of extending the hook's existing `refetchFallback`, which already runs on foreground and socket drop |
| **The cheapest layer was missing entirely** — the gate returns 403 with `code: terms_not_accepted`, and `baseApi` only handled 401 | Added the 403 branch; it turns the refusal itself into the prompt, with no Reverb involved |
| A second hook for the account channel would open a second websocket per session | Both channels subscribed from the one Echo instance |
| Broadcast should hang off activation, not publication | Moved to `onTermsVersionActivated`, with the reasoning in the docblock |
| Don't overload `compartment-status`; one event carrying `list<int> $recipientUserIds` so a chunk is one dispatch; payload a bare signal | All three as described |
| The planned check on whether the existing notification fires at the wrong moment was moot — publish and activate happen in one aggregate write | Item dropped |

### Implementation

| Finding | Applied |
|---|---|
| **`composer format-check` failed** — an import was out of order | Fixed; 360 files pass |
| **No mobile tests**, though the plan listed them | `isTermsNotAcceptedError` extracted into `termsGate.ts` and tested; `accountChannel.test.ts` pins the channel and event names. 19 → 26 tests |
| Spinner tracked the compartments query only, then tracked both — so a broadcast produced a spinner nobody pulled | Bound to the gesture instead, via local state cleared when both refetches settle |
| Import grouping in `baseApi.ts`; ADR dated the 24th | Both fixed |
| ADR missing | Written as ADR-0056 after some back and forth — the decision worth recording is not "use Reverb" (ADR-0023 settled that) but that account state gets its own channel rather than the convenient one |
| Two chunked passes over the user table | Left as is: notification hangs off publication and the broadcast off activation, so they are different events now and cannot share a walk without re-coupling what the review asked to separate |

### Found by running it, not by review

The first end-to-end attempt showed nothing in the app. A direct broadcast from tinker reached it immediately, which isolated the fault to the reactor path — the **event-worker was holding the pre-change reactor in memory**. Restarting the workers fixed it. Same failure mode as #227's credential issuance, and the reason a worker restart belongs in the deploy notes for both.

PR #275, opened 2026-09-13
build: make Just tasks cross-platform

### LINKED ISSUE #273: Make Just tasks work natively on Windows
## Problem

The root `Justfile` assumes a Unix shell for recipes that perform MQTT setup and tracing orchestration. On a stock Windows environment this causes commands to fail before their actual work begins:

- Embedded Bash/shebang recipes can trigger a missing `cygpath` lookup and effectively require Git Bash or another Unix compatibility layer.
- Tracing configuration evaluates `HOME`, which is not guaranteed to exist in Windows shells.
- The development app health check allows only five attempts with a five-second timeout and no startup grace period. Laravel initialization through Docker Desktop bind mounts takes considerably longer, so `just start` and `just trace-overlay` can fail with `locker-backend-app-1` marked unhealthy even though the app eventually starts.

Windows contributors should be able to use the same public Just commands as Linux and macOS contributors without installing a second shell environment.

## Expected behavior

- Simple recipes remain readable in the root `Justfile`.
- Complex MQTT and tracing tasks use native PowerShell on Windows and Bash on Unix-like systems while preserving equivalent inputs, defaults, output, mutations, and failures.
- Scripts derive repository paths independently of the caller's working directory and do not expose the MQTT password.
- Docker Desktop startup has enough health-check grace and request time to accommodate Windows bind-mount performance.
- CI validates Just parsing and platform dispatch on Windows and Ubuntu, then statically analyzes both PowerShell and Bash entry points.
- Repository line-ending rules prevent Windows checkout settings from breaking Unix shell entry points.

## Acceptance criteria

- `just --fmt --check`, `just --list`, and dry-runs of every recipe pass on Windows and Ubuntu runners.
- `just start` completes from a cold Docker Desktop stack and starts dependent workers.
- `just trace-overlay` completes and reports the backend as instrumented.
- PowerShell scripts pass PSScriptAnalyzer.
- Bash scripts pass ShellCheck.
- The setup guide documents the required Just version and Windows installation path.

issdn:
## Summary

Make the root task runner usable from stock Windows and Unix-like environments without requiring Windows contributors to install Bash, Git Bash utilities, or `cygpath`.

The previous `Justfile` embedded Bash implementations directly in recipes and evaluated Unix-specific environment such as `HOME`. On Windows this caused recipe startup to fail before the underlying task ran. The new structure keeps the `Justfile` as the stable, user-facing command menu and dispatches complex work to an OS-native PowerShell or Bash entry point.

Existing commands and intent remain the same: contributors still use `just setup-mqtt`, `just trace-up`, `just trace-down`, and `just trace-status` on either platform.

### Keep orchestration in `scripts/`

Short, declarative commands such as `start`, `stop`, `status`, `sim`, and `install-hooks` remain in the `Justfile`. MQTT configuration and tracing contain branching, environment handling, file generation, retries, cleanup, and error propagation, so those implementations now live under `scripts/`.

Executable repository automation lives under the conventional `scripts/` directory, separate from application code and reusable helper modules. Each complex task has paired `.ps1` and `.sh` implementations with the same inputs, defaults, mutation order, output, and failure behavior:

- Windows recipes invoke Windows PowerShell 5.1 directly.
- Unix recipes invoke Bash directly.
- Paths are derived from each script's location, so commands do not depend on the caller's working directory.
- MQTT configuration is written through a temporary file, does not print the password, and propagates Docker failures.
- Tracing keeps its existing SigNoZ version, port overrides, readiness polling, and retained-data behavior.

The `Justfile` disables implicit dotenv loading, selects an explicit shell per operating system, and requires Just 1.56 because top-level platform attributes depend on that version. Comments in the `Justfile` and paired scripts record the maintenance contract close to the code instead of keeping a separate research document that would become stale.

### Add task-runner CI

The new `Task Runner Quality` workflow catches platform-specific regressions that are easy to miss when contributors only develop on one operating system:

- An Ubuntu/Windows matrix checks formatting, lists the public recipes, and dry-runs every recipe to verify that Just parses the file and selects the correct platform command without requiring secrets or running services.
- ShellCheck validates the Bash implementations using a pinned container image.
- PSScriptAnalyzer validates the PowerShell implementations using a pinned module version.
- Path filters avoid running the workflow for unrelated application changes, while concurrency cancellation avoids wasting CI time on superseded commits.

### Enforce stable line endings

`.gitattributes` forces LF endings for the `Justfile`, Bash scripts, and PowerShell scripts. This prevents Windows checkout settings from adding carriage returns that can break Unix shebangs or shell parsing. PowerShell supports LF, so one canonical representation works on both platforms and also keeps generated files and diffs deterministic.

`Setup.md` now documents the minimum Just version, links to the official installation instructions for Windows and Linux, and describes the actual MQTT password precedence.

### Tolerate slower Docker Desktop startup

The development app health check now has a three-minute startup grace period and uses a 30-second interval and timeout. Laravel cache warmup and framework requests are substantially slower through Docker Desktop bind mounts; the previous five-second timeout and immediate five-attempt failure window marked a working app unhealthy before it could finish starting. Three consecutive failures after startup still detect an unavailable app, while normal Windows startup no longer blocks dependent workers.

## Windows behavior and known limitation

- Windows parsing and dry-runs now select the PowerShell entry points without the previous `cygpath` lookup or missing-`HOME` failure.
- The collector check in `trace-status` still uses a broad `name=otel-collector` filter, so a running SigNoZ collector can potentially be mistaken for the Open Locker collector. That behavior predates this structural refactor and remains outside its scope.

## Testing

Passed:

- `git diff --check`
- `just --fmt --check` on Windows
- `just --list` and dry-runs of every recipe on Windows
- PowerShell parser validation for both `.ps1` scripts
- PSScriptAnalyzer 1.25.0 for both `.ps1` scripts
- ShellCheck 0.11.0 for both `.sh` scripts
- Cold `just start` on Docker Desktop; the full stack became healthy and started in 163.5 seconds
- `just trace-overlay` on Docker Desktop; the instrumented stack became healthy and started in 147.6 seconds
- Delayed health verification reported `locker-backend-app-1` healthy with zero consecutive failures
- `just trace-status` reported the SigNoZ UI, Open Locker collector, and backend instrumentation as active

Pending:

- GitHub Actions checks are running on the pull request

Fixes #273


Justfile
```diff
@@ -1,152 +1,86 @@
...
-setup-mqtt:
-    #!/usr/bin/env bash
-    set -e
-    
-    # Load variables from .env if present
-    if [ -f locker-backend/.env ]; then
-        export $(grep -v '^#' locker-backend/.env | xargs)
-    fi
+# This file is the public command menu: keep portable one-command tasks here.
+# Put branching, retries, file generation, and platform logic in scripts/.
```
merona-merlin, line 4: Could we avoid maintaining complete Bash and PowerShell implementations of the same workflows? Just should remain the thin, discoverable command interface, while complex behavior is implemented once in a cross-platform runtime. Since this repository already uses Node/pnpm, a dependency-free `.mjs` implementation invoked by the same recipes on every platform would substantially reduce this change and prevent the two implementations from drifting.

Justfile
```diff
@@ -1,152 +1,86 @@
...
-# Backend-agnostic: exports to whatever locker-backend/docker/otel-collector.yml
-# targets. Use this when you run your own Jaeger/Tempo/Honeycomb — point the
-# collector at it first, then bring the overlay up.
-#
-# Start the stack with tracing on (bring your own trace backend)
+    git config core.hooksPath .githooks
+
+# Generate Mosquitto configuration from the local backend .env file.
+[windows]
+setup-mqtt:
```
merona-merlin, line 45: Could we remove the host-side `setup-mqtt` workflow entirely and generate the Mosquitto configuration when the container starts? A single POSIX script executed inside the Linux container would work identically on every host, eliminate both setup scripts, and make `docker compose up -d` sufficient. Please preserve the image’s existing Tini entrypoint by overriding only `command`, validate or escape the URL secret, and apply `umask 077` before writing the temporary configuration.

locker-backend/docker-compose.yml
```diff
@@ -77,9 +77,13 @@ services:
                 condition: service_started
         healthcheck:
             test: ["CMD", "curl", "-f", "http://localhost:8080/up"]
-            interval: 5s
-            timeout: 5s
-            retries: 5
+            # Laravel cache warmup is considerably slower through Docker
+            # Desktop bind mounts. Ignore failures while the entrypoint runs,
+            # then allow enough time for a full framework request.
+            start_period: 3m
```
merona-merlin, line 83: This health-check policy is independent of the cross-platform task-runner change and introduces a significant operability trade-off. Failures do not count for three minutes, after which the 30-second interval and timeout can keep dependent services blocked for several additional minutes. Could we move this into a focused change with its own rationale and ADR? If a long Docker Desktop grace period is required, `start_interval` could still probe frequently during startup without weakening normal failure detection.

.github/workflows/task-runner-ci.yml
```diff
@@ -0,0 +1,95 @@
...
+        uses: actions/checkout@v7
+
+      - name: Install Just
+        uses: extractions/setup-just@v4
+        with:
+          just-version: '1.58.0'
+
+      - name: Validate Justfile
+        run: |
+          just --fmt --check
```
merona-merlin, line 50: The matrix currently proves that Just renders the expected platform command and that both scripts pass static analysis, but it does not execute their behavior or prove that the Bash and PowerShell implementations remain equivalent. With one shared implementation, could we add `node:test` coverage for environment precedence, template generation, generated Docker commands, and tracing readiness behavior, then run those same tests on Windows and Ubuntu?

merona-merlin (changes requested):
The portable Just command surface is useful and should remain, but the paired shell implementations create a large permanent maintenance burden. Please keep Just as a thin command menu, move complex behavior into one cross-platform implementation, generate Mosquitto configuration inside the container, and separate the unrelated health-check policy. The new Just 1.56 requirement is also currently undocumented; remove it if the conditional Just features are no longer needed, or document it as required by the issue. Infrastructure, secret-handling, and health-check decisions should be captured in an ADR.

PR #278, opened 2026-09-13
fix(backend): repair provisioning token copying

### LINKED ISSUE #253: New provisioning token copy button not working
For me, when clicking the copy symbol when creating a new provisioning button, it didn't get copied to clipboard. I'm using chrome.  Not sure if this is a me-issue or a general issue, so someone should probably test this on another machine first.

issdn:
## Summary

Fix provisioning-token copying in the Filament admin modal. Previously, the copy action embedded the token in an Alpine click expression and could fail to copy the displayed token in Chrome. The modal now passes the one-time token to a dedicated browser component, copies that exact value through the Clipboard API, and shows a useful Filament error toast when browser clipboard access is blocked. The copy button changes to a check icon after success and resets when the user leaves and re-enters the control.

The change also registers the component as a Filament asset, keeps the token out of Livewire action state and snapshots, and adds regression assertions for rendering, asset registration, icon state, and error handling.

## Implementation

The browser component owns the copy state and receives translated error text from Blade. This keeps the template focused on layout and lets the Clipboard API failure path use the same Filament notification system as the rest of the admin panel. Manual copying remains the fallback when browser permissions or context prevent clipboard access.

## Testing

- `node --check resources/js/filament/provisioning-token-copy.js`
- `node --check public/js/app/provisioning-token-copy.js`
- Laravel Pint passed for the changed Blade and test files.
- `php artisan test --filter=test_filament_edit_issue_shows_returned_token_only_once --compact` passed (31 assertions).

Fixes #253


locker-backend/resources/js/filament/provisioning-token-copy.js
```diff
@@ -0,0 +1,22 @@
+window.provisioningTokenCopy = (token, messages) => ({
+    token,
+    copied: false,
+
+    async copyToken() {
+        try {
+            await navigator.clipboard.writeText(this.token);
```
merona-merlin, line 7: Can we first verify the security context in which the original failure occurred? The Clipboard API requires HTTPS, although `http://localhost` is treated as secure by Chrome. If the page was opened through a LAN IP or another plain-HTTP hostname, this implementation will still be unable to copy and will only show the fallback notification. Please reproduce the issue on `http://localhost` or HTTPS and document the required development URL, or describe this change as error handling rather than a clipboard fix.

locker-backend/resources/views/filament/actions/show-provisioning-token.blade.php
```diff
@@ -1,36 +1,45 @@
...
-                })
-            "
-        />
+    <div class="fi-input-wrp">
+        <div
+            class="fi-input-wrp-content-ctn"
+            style="padding-block: 1rem; padding-inline: 1.25rem;"
+        >
+            <code
+                class="fi-sc-text fi-font-mono"
```
merona-merlin, line 14: Could we retain `select-all` and `break-all` here? The error message explicitly asks users to select and copy the token manually when clipboard access is blocked. Removing these classes makes that fallback harder and may cause the 64-character token to overflow on narrow layouts.

locker-backend/resources/views/filament/actions/show-provisioning-token.blade.php
```diff
@@ -1,36 +1,45 @@
...
+                data-testid="one-time-provisioning-token"
+            >{{ $token }}</code>
+        </div>
+        <div
+            class="fi-input-wrp-suffix"
+            x-on:mouseenter="resetCopyState()"
```
merona-merlin, line 20: Could we avoid resetting this state only on `mouseenter` and replacing the focused button with a second button? Touch and keyboard users may never trigger the reset, and hiding the focused button can cause focus loss. The previous `role="status"` announcement has also been removed. A single button with a dynamic icon/label plus an `aria-live` status, or the previous timeout-based reset, would preserve keyboard and screen-reader behavior.

locker-backend/tests/Feature/LockerBankProvisioningResetTest.php
```diff
@@ -393,7 +394,36 @@ public function test_filament_edit_issue_shows_returned_token_only_once(): void
 
         $modalContent = $component->instance()->getMountedAction()?->getModalContent();
         $this->assertNotNull($modalContent);
-        $this->assertStringContainsString($token, $modalContent->render());
+        $renderedModalContent = $modalContent->render();
+        $this->assertStringContainsString($token, $renderedModalContent);
```
merona-merlin, line 398: Non-blocking: these assertions verify JavaScript source strings rather than the clipboard behavior that originally failed in Chrome. Could we either cover the component with a small JavaScript test using a mocked Clipboard API, or document manual Chrome verification for both successful and blocked clipboard access?

merona-merlin (changes requested):
The extraction of the clipboard logic is a good direction, and keeping the token out of Livewire state remains well covered. Before merging, please verify the original failure's security context, restore a usable manual-copy fallback, and preserve keyboard and screen-reader behavior for the copied state. The branch is also 28 commits behind `main`; please update it so the currently failing backend image security check is rerun against the fixed dependencies.

locker-backend/resources/js/filament/provisioning-token-copy.js
issdn: Doesn't work on localhost (secure context) nor over HTTPS.

Issue #266, opened 2026-09-01
Discuss dropping the `dev` branch in favor of trunk-based development


merona-merlin:
## Context

This is a **discussion**, not an implementation ticket. Please comment with opinions, constraints, and preferred options. No workflow or branch change until we decide.

Today we follow [ADR-0043](https://github.com/Open-Locker/Open-Locker/blob/main/docs/adr/0043-monorepo-release-strategy.md):

- Day-to-day work lands on `dev` via PRs.
- When a set of changes is ready to ship, `dev` is merged into `main`.
- Releases are **tags** on `main` (`backend-v*`, `client-v*`, `mobile-v*`), not the merge itself.
- Hotfixes may land on `main` and must be merged back to `dev`.

The GitHub default branch is already `main`. CI runs on both `main` and `dev`. Website CD is `main` only. Docker `latest` is gated to `main`; `dev` can publish a separate image channel. Mobile tracks in the ADR map `dev` → off-store internals, `main` → store beta, tags → versioned release.

Related (closed): #50 (release strategy), #102 (dev-branch CI).

**Question:** should we drop the long-lived `dev` branch and do trunk-based development on `main`?

## What "trunk-based" would mean here

- `main` is the only long-lived branch.
- Feature work uses short-lived branches and PRs **into `main`**.
- Trunk stays releasable; unfinished product work is isolated with flags, drafts, or not merged yet.
- Releases stay tag-driven (this already matches ADR-0043 decision 4).
- A change like this would **supersede** ADR-0043 decision 3 (and the `dev`/`main` channel mapping in decision 7), not silently rewrite that ADR.

## Why this is on the table

- `dev` → `main` is extra ceremony, and the two branches can drift (the ADR already requires hotfix back-merges).
- The ADR rollout notes that `main` has already trailed `dev` by a large stack — the integration branch can become the real source of truth while `main` goes stale.
- We already do not treat "merge to `main`" as a SemVer release. Tags do that. That is closer to trunk-based than to Git Flow.
- Website already deploys from `main` only, so work that lives only on `dev` never reaches the public site until a promotion merge.
- Small team / OSS: two long-lived branches is easy to get wrong (PRs against the wrong base, stale `main`, forgotten promotion).

## Option A — Keep `dev` + `main` (status quo)

**Pros**

- Unfinished integration stays off the production-adjacent branch.
- Clear promotion step: "this set of `dev` commits is ready for `main`."
[...]

PR #280, opened 2026-09-15
docs: adopt trunk-based development on main

Aibolit66:
Records the decision taken in #266: drop the long-lived `dev` branch and develop on `main`.

## What the ADR decides

1. `main` is the only long-lived branch; short-lived branches open PRs into `main`.
2. Releases stay tag-driven — ADR-0043 decision 4 is unchanged.
3. Channel mapping: website CD and both Docker `latest` tags follow `main` tip; the locker-client `:dev` image tag is retired; mobile is already tag-only per ADR-0055.
4. Fielded devices pin `client-v*` / `backend-v*` tags and must never track `latest`. This is what keeps a merge to `main` from reaching hardware.
5. Trunk stays releasable through required status checks and author verification. Approving reviews are **not** a merge gate: the team has no separate QA function, so the author verifies their own change.

Supersedes ADR-0043 decision 3 and the `dev` parts of decision 7. The rest of ADR-0043 — tag namespaces, SemVer policy, tag-driven releases — stays in force. ADR-0043 itself is not rewritten.

## Notes

- `docs/adr/0052` does not exist; this is numbered 0058 after 0057.
- Alternative D records why a mandatory approving review was rejected, so the reasoning survives if we revisit it.
- Implementation (workflow triggers, process docs) is in a separate PR.

Tracks #266.


Aibolit66:
Folding this into #281 — one ticket, one PR. The ADR commit now sits on `chore/266-retarget-ci-to-main` alongside the workflow and docs changes.

Issue #249, opened 2026-08-24
Audit log missing events


CrazyCowLady:
None of my attempts to open a locker before accepting the terms are showing in the audit log.
The log shows: new terms created by admin > terms accepted by user > locker opened by user. 
What I actually did: new terms created by admin > user tried to open, but denied (2x) > terms accepted by user > locker opened by user.

Also:
When editing users, you can see the "last open status" of their compartments, but this also doesn't show whether a user tried and failed to open a compartment and got a 403. I assume this stems from the same issue.

Aibolit66:
PR #294

Issue #202, opened 2026-08-11
Should every stored event require an audit-log classification?


merona-merlin:
## Question

Should every new class under `app/StorableEvents` be explicitly classified as either visible in the admin audit log or intentionally excluded?

## Context

The audit log currently uses a curated whitelist so high-volume telemetry does not drown out meaningful domain actions. The downside is that a new business-relevant event can silently remain absent when its whitelist entry and presentation are forgotten.

This happened with `GroupArchived` while reviewing #181. ADR-0026 already identifies whitelist drift as a risk, but currently relies on code review to prevent it.

## Proposal for discussion

Introduce an explicit classification contract:

- Every stored event is classified centrally as either **auditable** or **excluded**.
- Excluded events include a short reason, such as high-volume telemetry or internal transport traffic.
- A test discovers all stored event classes and fails when an event is unclassified or appears in both groups.
- Auditable events retain a generic fallback label/description, while important events can provide a tailored human-readable presentation.
- If adopted, document the decision in ADR-0026 or a superseding ADR.

The intent is not to automatically expose all events. It is to make adding a stored event require a conscious audit-log decision.

### Benefits

- Prevents relevant events from silently disappearing from the audit log.
- Keeps telemetry exclusions explicit and reviewable.
- Makes the audit policy enforceable in CI rather than relying only on reviewer memory.

### Costs and risks

- Every new event requires an additional classification step.
- Runtime class discovery would add unnecessary complexity; discovery should happen only in tests.
- A generic fallback may be technically complete but still insufficiently clear for administrators.

## Alternatives

1. Keep the current whitelist and rely on review discipline.
2. Show every event except an exclusion denylist; simpler coverage, but greater risk of noisy or sensitive events appearing unexpectedly.
3. Use attributes or marker interfaces on event classes instead of a central registry.

## Discussion

- Is CI enforcement appropriate, or is the current review process sufficient?
- Should classification live in a central registry, attributes, or marker interfaces?
- Should an auditable event require a custom human-readable description, or is a generic fallback acceptable initially?

merona-merlin:
I would especially appreciate opinions from others working on event-sourced flows and the admin UI: does this classification contract solve a real maintenance problem, or does it add too much ceremony for each new event? Alternative designs are welcome before we turn this into an implementation task.

merona-merlin:
@CrazyCowLady @Aibolit66

Aibolit66:
@merona-merlin  That's a good idea I guess.
[...]

PR #294, opened 2026-09-17
fix(backend): record open refusals that middleware used to swallow (#249)

Aibolit66:
Fixes the missing audit entries reported in #249.

## The bug

Opening a locker before accepting the terms left no trace anywhere. `RequireAcceptedTerms` and `EnsureVerifiedEmailApi` answered 403 before `CompartmentController::open` ran, so no open attempt was ever recorded — nothing in the audit log, nothing in the user's open history, and "Last open status" on the user edit screen still read "No requests" after a user had been refused twice. Both halves of the report share that one cause.

## The fix

The open route decides both refusals itself. `CompartmentAccessService::requestOpen` records `CompartmentOpenRequested`, then denies with `terms_not_accepted` or `unverified_email`. The existing projector writes the refusal to `compartment_open_requests` and the existing whitelist surfaces it in the audit log — no new event class, no migration, no whitelist change. The `unverified_email` branch was already there and unreachable; it works now.

Refusal bodies are unchanged and single-sourced from the middleware classes, so the `terms_not_accepted` body the mobile app keys on (`mobile-app/src/store/termsGate.ts`, the #250 fix) cannot drift. The route is excluded from the terms middleware by name; `route:list` confirms every other route in the group keeps it.

Terms enforcement is a required argument rather than a default, so each caller states its policy and PHPStan catches an omission. The Filament panel opts out and behaves exactly as on `main`: it offers no way to accept terms, so gating there would strip the Open button from the admin who just published a version, with no in-panel remedy. That asymmetry is the subject of `docs/adr/0060-open-refusals-decided-inside-the-open-path.md`.

## Verified locally

Reproduced the reported scenario against the local stack: activated a terms version, refused an open twice as a normal user, and confirmed the audit log now shows `Open requested` → `Open denied (terms_not_accepted)` for each attempt, where `main` shows nothing.

- `composer test` — 433 tests, 1611 assertions, zero failures. 3 errors are pre-existing on `main` (`LockerBankProvisioningResetTest` via the untouched `InteractsWithOneTimeProvisioningToken.php:40`).
- `composer format-check` — PASS, 366 files.
- `composer analyse` — 2 errors, both that same pre-existing file. Nothing new.

New coverage: terms denial recorded, unverified denial recorded, terms outranks unverified, and the amended unverified case in `CompartmentAccessControllerTest` that previously asserted the bug. Three further tests are regression guards that also pass on `main` — they pin the panel path, the sibling routes' gating, and the version-rollback predicate against future drift.

## Notes for the reviewer

- **Deploy**: the middleware change is inert until routes are re-cached. Needs `route:cache` or a cleared cache.
- The denial reason renders as raw `terms_not_accepted` in the audit log, matching how the pre-existing `missing_active_access` already renders. Prettifying reasons is #202's territory and deliberately out of scope here.
- Scramble's `@response` docblocks turn out to be decorative — the tag is not read by the version in use, on `main` as much as here, so the 403 shapes are absent from the generated spec. The attribute form Scramble does read collapses both 403s into one and types the body as a string, which is worse than documenting nothing. Doing it properly needs a `oneOf` schema and a client regeneration, so it belongs in its own change.


merona-merlin:
@Aibolit66 can you please check why the tests are failing?

Aibolit66:
image

@merona-merlin 
I don't see anything failing
what do you mean?

